PR #4, opened 2021-02-09
Build for multiple environments

rsbondi:
I am admittedly not well versed in this but it seems to work, I just copied what I needed from the default rollup lib template.  It builds 3 files to `/dist` that will choose the correct one for the environment.  I was able to run in a node environment and in the browser with `nostr-web-client-poc`.  This of course will require the build step before publishing to npm.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,9 @@
   "name": "nostr-tools",
   "version": "0.3.0",
   "description": "Tools for making a Nostr client.",
-  "main": "index.js",
+  "main": "dist/nostr-tools.cjs.js",
+  "module": "dist/nostr-tools.esm.js",
+  "browser": "dist/nostr-tools.umd.js",
   "repository": {
     "type": "git",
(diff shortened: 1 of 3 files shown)


rsbondi:
ok, it seems there is still an issue with the `WebSocket` is native in browser but is a package in node so this is not ready yet.

fiatjaf:
I think I was using a rollup build step here before but it turned out nothing worked as I wanted so I removed it and left that burden for the programs that dependended on this to "build" it.

To fix the websocket stuff I believe you must install the `ws` package and add a `globals: {'WebSocket': 'ws'}` to the rollup config. I don't know if doing something on this library can help with that. Probably, but I have no idea.

rsbondi:
I was able to get the websocket working in node, but with inconsistent results in the browser.  I think I should just build just for node and point the browser to the non built index.js as before.  I will submit another if I can make satisfactory progress.

fiatjaf:
That sounds like a good idea.

PR #5, opened 2021-02-10
Add build for use in nodejs environment

rsbondi:
This builds as a node module leaving the `index.js` as the entry point otherwise.  `WebSocket` is native to the browser but the `ws` package is imported for node

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,0 +1,15 @@
+import pkg from './package.json';
+
+export default [
+	{
+		input: 'index.cjs.js',
+		external: [
+			"buffer",
+			"noble-secp256k1",
+			"ws"
+		],
+		output: [
+			{ file: pkg.main, format: 'cjs' }
+		]
+	}
+];
--- a/index.cjs.js
+++ b/index.cjs.js
@@ -1,0 +1,5 @@
+import WebSocket from 'ws'
+global.WebSocket = WebSocket
+import * as nostr from './index'
+
+export default nostr
--- a/package.json
+++ b/package.json
@@ -2,14 +2,20 @@
   "name": "nostr-tools",
   "version": "0.3.2",
   "description": "Tools for making a Nostr client.",
-  "main": "index.js",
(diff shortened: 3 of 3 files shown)


rollup.config.js
```diff
@@ -0,0 +1,11 @@
+import pkg from './package.json';
+
+export default [
+	{
+		input: 'index.js',
+		external: ['ms'],
```
fiatjaf, line 6: What is this 'ms'?

relay.js
```diff
@@ -1,4 +1,5 @@
 import {verifySignature} from './event'
+const WS = typeof window === 'undefined' ? require('ws') : WebSocket
```
fiatjaf, line 2: Does this require work when building for the browser? I think any transpiler will always automatically include the `ws` package since it will be available.

rsbondi: there is no build for the browser, see `package.json`

```
  "main": "dist/nostr-tools.cjs.js",
  "module": "index.js",
  "browser": "index.js",
```

rollup.config.js
rsbondi: It was in the rollup template, may be specific to rollup, but don't really know what it is

rsbondi: I think I can delete, will sqaush out

relay.js
rsbondi: Well, I guess there is a build for anyone using the library, so their build would be bulky, don't know if there is a better way.

rsbondi: I added a separate `index.cjs.js` file which imports and exports the original, leaving `relay.js` untouched, this should eliminate the `ws` in the browser when building by third parties using `nostr-tools`

fiatjaf:
Going to the brute-force mode with https://github.com/fiatjaf/nostr-tools/commit/d355bbead48651ad2e304ff43409f8bd8bb4d0dd. I think that's what everybody does. Adding this polyfill for everybody. Let's hope the library authors know what they're doing. Should fix this.

PR #6, opened 2021-02-19
Correct subscriptions for new protocol

rsbondi:
I think most of these changes should be obvious, relay.js line 89, id does not exist so removed from log

The bind on the channel was only bound once, on return so it kept the same value for every subscription.  I moved it to `sub` and also added the reduce to convert to hex

And `trySend` already stringifies, so it was removed from `realy.publish`

### DIFF
--- a/relay.js
+++ b/relay.js
@@ -115,13 +115,13 @@ export function relayConnect(url, onNotice) {
     }
   }
 
-  const sub = async ({cb, filter}) => {
-    const channel = (await sha256(Math.random().toString())).reduce((str, byte) => str + byte.toString(16).padStart(2, '0'), '')
+  const sub = async ({ch, cb, filter}) => {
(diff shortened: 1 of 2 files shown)


relay.js
```diff
@@ -116,7 +115,8 @@ export function relayConnect(url, onNotice) {
     }
   }
 
-  const sub = async (channel, {cb, filter}) => {
+  const sub = async ({cb, filter}) => {
+    const channel = (await sha256(Math.random().toString())).reduce((str, byte) => str + byte.toString(16).padStart(2, '0'), '')
```
fiatjaf, line 119: There is a problem here. Only the first call should get a random channel ID. The following calls derived from this should reuse that ID.

fiatjaf: You see how each `sub()` call returns an object with `.sub()`? The idea is that calling that sub-`sub` function should make a new `REQ` call to the relay with the same previous `channel` id.

This is confusing.

rsbondi: a25e6c4265374bf5448105d89439004a86132014

PR #7, opened 2021-02-21
Return relay with sub callback for new relay

rsbondi:
I don't quite understand well enough but this is what I observe to be occurring
[here](https://github.com/rsbondi/nostr-tools/blob/6750a6cc7659a58cd8cc646c7001aa2ead98c977/pool.js#L35)(this commit) will catch it the first time when the relay is first added, and [here](https://github.com/rsbondi/nostr-tools/blob/6750a6cc7659a58cd8cc646c7001aa2ead98c977/pool.js#L24) on subscriptions that are added after

So if I subscribe, then add an additional relay, it is covered by this PR

### DIFF
--- a/pool.js
+++ b/pool.js
@@ -32,7 +32,7 @@ export function relayPool(globalPrivateKey) {
           sub(id, {cb, filter})
         ]),
       addRelay: relay => {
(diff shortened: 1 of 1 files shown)


monlovesmango:
I think the current version already accounts for this now. should this be closed?

PR #11, opened 2022-04-06
import 'Buffer'

monlovesmango:
'Buffer' wasn't imported initially and was causing issues when I tried to use generatePrivateKey in a client I am building. not sure why Branle has no error, maybe I am doing something wrong?


fiatjaf:
These JS bundlers these days have ways to inject "Node dependencies" via some option or plugin and pretty much everybody does it. `Buffer` is a global on Node.

PR #13, opened 2022-05-05
query all users for particular domain

monlovesmango:
I think it would be helpful to add a function to query all the names for a particular domain, so that you can explore users on a particular domain. easy enough to implement on client side but might be good have in the base tools (feel free to disagree).

### DIFF
--- a/nip05.js
+++ b/nip05.js
@@ -26,3 +26,17 @@ export async function queryName(fullname) {
     return null
   }
 }
+
+export async function queryNames(domain) {
+  try {
+    if (!domain) return null
+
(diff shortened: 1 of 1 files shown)


fiatjaf:
Isn't this the same as the [`searchDomain`](https://github.com/monlovesmango/nostr-tools/blob/31ba86eee0a09684c1c85ba0b13f9581be36b017/nip05.js#L3) function?

monlovesmango:
lol yes it is. sorry got blinders on. thank you for pointing me to this!!

PR #15, opened 2022-06-22
update cb property in subControllers relay entry

monlovesmango:
when updating subscription or adding new relays, subsequent events that are received have the relay as undefined. by updating cb property for the subControllers entries to be an arrow function (when calling sub.sub or sub.addRelay), subsequent events now return the relay url appropriately.

I also rearranged order in line 37 so that order is consistent throughout.


fiatjaf:
Published.

PR #16, opened 2022-07-02
NPM build and publish

Dolu89:
PR for publishing ESM module to NPM unders ./dist/nostr.js


fiatjaf:
Is this working? Will it break things?

PR #20, opened 2022-09-06
fix: CJK characters are garbled after decryption

bieberg0n:
I'm from Asia and I found that CJK characters are garbled after decryption.
Because during decryption, parameters of `decipher.update` does not contain `outputEncoding`.

Test code:
```Javscript
import { decrypt, encrypt } from './nip04.js'

const main = function() {
  let priv = '3277614521b0dba8ec930e64ff4ea5bb2709dfd67d35586b8b380d0beeb420c1'
  let pub = '8643c759c757bbaf06eae40ac4454627ec44731a7f48a4ba04244b68fd028999'
  let txt = '苟利国家生死以，岂因祸福避趋之'
  let cipText = encrypt(priv, pub, txt)
  let r = decrypt(priv, pub, cipText)
  console.log(r)
}

main()
```
Result:
image

After repair:
image

### DIFF
--- a/nip04.js
+++ b/nip04.js
@@ -29,7 +29,7 @@ export function decrypt(privkey, pubkey, ciphertext) {
     Buffer.from(normalizedKey, 'hex'),
     Buffer.from(iv, 'base64')
   )
-  let decryptedMessage = decipher.update(cip, 'base64')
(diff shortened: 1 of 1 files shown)


fiatjaf:
Good catch! Thank you very much.

fiatjaf:
Released in 0.24.1.

PR #24, opened 2022-10-04
modify ./nip04.js to use buffer npm package

snuffyDev:
This will use the npm package instead of node's core module. 

The README for the buffer package from NPM (https://github.com/feross/buffer#usage) mentions using `'buffer/'` instead of `'buffer'`, since `'buffer'` will resolve to Node's Buffer module.

### DIFF
--- a/event.js
+++ b/event.js
@@ -1,58 +1,49 @@
-import {Buffer} from 'buffer'
-import createHash from 'create-hash'
-import * as secp256k1 from '@noble/secp256k1'
+import { Buffer } from "buffer/";
+import createHash from "create-hash";
+import * as secp256k1 from "@noble/secp256k1";
 
 export function getBlankEvent() {
-  return {
-    kind: 255,
-    pubkey: null,
-    content: '',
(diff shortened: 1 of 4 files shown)


fiatjaf:
Thank you for the PR, but why do we want to use the npm package instead of Node's Buffer module?

snuffyDev:
> Thank you for the PR, but why do we want to use the npm package instead of Node's Buffer module?

It’s installed as a dependency: https://github.com/fiatjaf/nostr-tools/blob/1f1bcff8035a3e2926bdc8a38a5e6a90222dcec4/package.json#L14

But also for usage in a browser. If it’s preferred to use the Node module in a Node environment, I can add a check

fiatjaf:
It's installed as a dependency because it's needed in the browser. I don't understand JavaScript enough to tell if it is better to use npm dependency on Node.js too, that's why I am asking.

If you add a check wouldn't that be equivalent to what we have today without this PR?

fiatjaf:
:(

PR #26, opened 2022-10-24
replace micro-bip with @scure/bip, fix #25

### LINKED ISSUE #25: replace micro-bip32/39 with @scure/bip32/39
according https://www.npmjs.com/package/micro-bip39

they suggest to switch to @scure/bip39 which with the security audit

same for bip32 https://www.npmjs.com/package/micro-bip32

gasolin:
update packages described in #25


fiatjaf:
Thank you!

PR #28, opened 2022-11-24
Filter type with optional atributtes

arejula27:
When using typscript is really annoying to create the whole Filter type, with this change it will be more comfortable use a filer. The next image shows  an example
image


fiatjaf:
Good catch, I hadn't realized this because I don't use TypeScript.

arejula27:
Do u mind if I do more pr on similar cases when I find them?

fiatjaf:
No, please send PRs!

PR #30, opened 2022-11-25
Documentation pool.js

arejula27:
I spend a time studying the pool.js , at the same time i make some anotations.Maybe coud be a good option to add it.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -32,13 +32,13 @@ declare function signEvent(event: Event, key: PrivKey): Promise<[Uint8Array, num
 
 // filter.js
 declare type Filter = {
-    ids: string[],
-    kinds: EventKind[],
-    authors: string[],
-    since: number,
-    until: number,
-    "#e": string[],
-    "#p": string[],
+    ids?: string[],
+    kinds?: EventKind[],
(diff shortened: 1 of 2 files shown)


arejula27:
I noticed that when you publish a event never rejects, so it will be blocked. It is a correct behaviour?

fiatjaf:
In theory yes, but we could put a timeout there.

And we must also add support for https://github.com/nostr-protocol/nips/blob/master/20.md, so we can get an `OK` (or not ok) response for each event published.

arejula27:
i could try it, where can i see the status codes?

fiatjaf:
What status codes? It's either `["OK", <id>, true]` or `["OK", <id>, false]` and then a string message. Some relays will return that right after you send an event.

arejula27:
image

That status, o dodnt dont understand what does it mean. 

As i see the implementation of the nip 2o will be on the line 116 of the relay.js, so i dont understant what the previous status means. Also could be interesting do a publish sincronous and asyncrous. Im currently studing diffents approaches

arejula27:
Can this pr be merged?

Issue #32, opened 2022-11-26
Implement nip 20


arejula27:
Add support for https://github.com/nostr-protocol/nips/blob/master/20.md, so we can get an OK (or not ok) response for each event published.

arejula27:
As I commented in the pr #30 would be a good idea to have two implementations, one synchronous and other asynchronous

arejula27:
We should discuss if we want to implement this nip with a call back function or more like a *wait until* when publishing and event. Opinions?

fiatjaf:
If you just return a Promise the caller can choose to use `await` or not. Let's not use callbacks if possible.

arejula27:
As i see the response will come from the web socket of line 75 in *relay.js", so i dont think is posible to handle it on the pusblish function of pool.js. Maybe i dont understand correctly?

arejula27:
i tthink i get it, pr later :D

fiatjaf:
You can create a Promise and return it, but keep a reference to the `resolve` function of that promise, then call it when you want.

(for future reference)

fiatjaf:
This is done with event listeners.

PR #36, opened 2022-11-27
Nip 20

arejula27:
I try to implement the nip 20  as we talked on issue #32  on the relay.js, please take a look. If its correct i will continue on the pool file

### DIFF
--- a/relay.js
+++ b/relay.js
@@ -118,9 +118,11 @@ export function relayConnect(url, onNotice = () => {}, onError = () => {}) {
           case 'OK':
             if (data.length !== 4) {
               // ignore malformed EVENT
+              return
+            }
+            if (pendingConfirmations[event.id]) {
               pendingConfirmations[event.id]()
               delete pendingConfirmations[event.id]
-              return
             }
         }
       }
--- a/relay.js
+++ b/relay.js
@@ -30,6 +30,7 @@ export function relayConnect(url, onNotice = () => {}, onError = () => {}) {
 
   var eventListeners = {}
   var eoseListeners = {}
+  var pendingConfirmations = {}
 
   function connect() {
     ws = new WebSocket(url)
@@ -114,6 +115,13 @@ export function relayConnect(url, onNotice = () => {}, onError = () => {}) {
               eventListeners[channel](event)
             }
             return
+          case 'OK':
+            if (data.length !== 4) {
+              // ignore malformed EVENT
+              pendingConfirmations[event.id]()
+              delete pendingConfirmations[event.id]
+              return
+            }
         }
       }
     }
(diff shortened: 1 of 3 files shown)


relay.js
```diff
@@ -114,6 +115,13 @@ export function relayConnect(url, onNotice = () => {}, onError = () => {}) {
               eventListeners[channel](event)
             }
             return
+          case 'OK':
+            if (data.length !== 4) {
+              // ignore malformed EVENT
+              pendingConfirmations[event.id]()
```
fiatjaf, line 121: must check if this pendingConfirmation exists

relay.js
```diff
@@ -114,6 +115,13 @@ export function relayConnect(url, onNotice = () => {}, onError = () => {}) {
               eventListeners[channel](event)
             }
             return
+          case 'OK':
+            if (data.length !== 4) {
```
fiatjaf, line 119: shouldn't this check be a `===`?

relay.js
```diff
@@ -125,6 +133,19 @@ export function relayConnect(url, onNotice = () => {}, onError = () => {}) {
     connect()
   } catch (err) {}
 
+  async function trySendWithConfirmation(params) {
+        trySend(params)
```
fiatjaf, line 137: there is a possible race condition here: since you call `trySend()` first it might be that the `OK` event will be received before you have the chance of registering `pendingConfirmations[event.id]`.

relay.js
```diff
@@ -178,6 +199,9 @@ export function relayConnect(url, onNotice = () => {}, onError = () => {}) {
   return {
     url,
     sub,
+    async publishWithConfirmation(event) {
+      return trySendWithConfirmation(['EVENT', event])
+    },
```
fiatjaf, line 204: I think it is better to nuke the `publish` method and replace it with this new approach that returns a Promise.

but there is a catch: the Promise cannot rely solely on the `OK` message. it must wait for the `OK` message, but if that doesn't happen it must do as the current `publish()` does: subscribe to the event and wait for it to be returned -- then if all of that fails after a while call the `reject` callback of the promise.

what do you think?

relay.js
arejula27: I didn't realize about that, totally agree. I will think about that

relay.js
arejula27: I really dont understand the first approach so i did the new one for dont messing it up. I think it is a good solution.
Can u explain me the first publish function? what are the callbacks and the status? What is expecting the sub created? only the same event? It is a kind of confirmation out of the nips? Maybe is write an event and then trying to read it for confirm?

arejula27:
Didn't have time to reimplemented, a pr will be on holiday 🫡

fiatjaf:
I'll do this myself, must refactor some stuff here.

PR #37, opened 2022-11-28
event fields

arejula27:
some event fields were missing


fiatjaf:
Thank you, but it is `sig`, not `signature`, right?
If it is `signature` anywhere else that place is wrong.

arejula27:
I didn't checked it, will correct later if it is wrong

arejula27:
You were right, fixed and merged.

PR #41, opened 2022-12-05
refactor of cb api

monlovesmango:
hey fiatjaf and all,

here is my initial refactor of the call back api as proposed by fiatjaf. it replaces all the existing call back arguments with listeners that can be added to the pool, subscription, or relay (if using relay module without the pool module). only 'event' and 'eose' can listen at the sub level, but these can also listen at the pool or relay (again, only if using relay module without the pool module) level as well. current list of listener types:

- event
- eose
- connection
- disconnection
- error
- notice

format of the calls:
pool.on('connection', cb)
pool.on('event', cb)
sub.on('eose', cb)
relay.on('notice', cb)

this hasn't been thoroughly tested yet (though seems to work from cursory testing) but wanted to get eyes on this sooner rather then later in case I need to be redirected.

other major changes: 

- I changed connectRelay to relayInit and then added a new method 'connect' to connect to the relay. I did this because I didn't know how else to allow creation of listeners on the relay prior the relay connecting. please let me know if there is a better method of solving this problem (I see a similar problem with creating the sub, where you don't have a chance to set listeners on the sub prior opening the sub).
- all listener call back functions take only 1 argument, an object that has the following keys:
   * relay: relay url
   * type: type of listener
   * id: sub id for sub specific listeners ('EVENT' or 'EOSE')
   * event: event object, only for 'event' listener
   * notice: notice message, only for 'notice' listener

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,41 +14,59 @@ pool.setPrivateKey('<hex>') // optional
 pool.addRelay('ws://some.relay.com', {read: true, write: true})
 pool.addRelay('ws://other.relay.cool', {read: true, write: true})
 
-// example callback function for a subscription
-function onEvent(event, relay) {
-  console.log(`got an event from ${relay.url} which is already validated.`, event)
+// example callback functions for listeners
+// callback functions take an object argument with following keys:
+//  - relay: relay url
+//  - type: type of listener
+//  - id: sub id for sub specific listeners ('EVENT' or 'EOSE')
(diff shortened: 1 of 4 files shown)


monlovesmango:
hey fiatjaf, this is ready for review now. I have done a bunch of unit tests and also plugged it into astral in dev env and everything seems to work smoothly.

PR #45, opened 2022-12-20
fixing import statements in examples

wds4:
If I am not mistaken, the examples in the README need to be changed from a require to an import statement.


fiatjaf:
Thank you, but I had fixed those already the minute before I saw the PR!

PR #67, opened 2022-12-25
Make `opts` arg optional for `sub` method

t4t5:
In the README and the code, it looks like the second argument for the relay's `sub` method is optional:

```typescript
let sub = relay.sub([
  {
    ids: ['d7dd5eb3ab747e16f8d0212d53032ea2a7cadef53837e5a6c66d42849fcb9027']
  }
])
```

In the type definitions it's required however, which leads to an error in editors:

CleanShot 2022-12-25 at 13 30 01@2x

Let's mark it as optional in the type definitions too! 👍

### DIFF
--- a/relay.ts
+++ b/relay.ts
@@ -8,7 +8,7 @@ export type Relay = {
   status: number
   connect: () => Promise<void>
   close: () => Promise<void>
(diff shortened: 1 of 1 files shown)


fiatjaf:
Thank you!

PR #77, opened 2023-01-03
Add link to nostr-relaypool-ts

adamritter:


fiatjaf:
Sorry for being harsh here, but I don't like this relaypool implementation, or any relaypool implementation at all. It is naïve and Nostr clients need smarter relay management to accommodate for a situation in which there will be many relays with different usersets, personal relays, restricted relays, niche topic relays and so on.

This is why I've removed the relay pool implementation nostr-tools have before.

So I would prefer to not advertise this as I consider it a bad practice and harmful to the network in the long run.

PR #89, opened 2023-01-16
refactor: use forEach

escapedcat:
Tiny refactor. Related tests are passing locally.


fiatjaf:
But why?

escapedcat:
Got it. Yeah, was just relying on the test there.

fiatjaf:
You would need 100% coverage before you could do that safely.

escapedcat:
Would you take a PR to cover the returns?

fiatjaf:
No, please, leave it like it is.

Issue #88, opened 2023-01-16
Unit-test: FAIL  ./nip04.test.js   ● encrypt and decrypt message


escapedcat:
1. Checking out master
2. yarn
3. yarn test

Gives me:
```sh
Test Suites: 1 failed, 7 passed, 8 total
Tests:       1 failed, 17 passed, 18 total
Snapshots:   0 total
Time:        4.634 s
```

Failing test:
```sh
 FAIL  ./nip04.test.js
  ● encrypt and decrypt message

    TypeError: Cannot read property 'importKey' of undefined

      392 |   let iv = Uint8Array.from((0, import_utils2.randomBytes)(16));
      393 |   let plaintext = utf8Encoder.encode(text);
    > 394 |   let cryptoKey = await crypto.subtle.importKey(
          |                                       ^
      395 |     "raw",
      396 |     normalizedKey,
      397 |     { name: "AES-CBC" },

      at Object.importKey (lib/nostr.cjs.js:394:39)
      at Object.encrypt (nip04.test.js:13:47)
```

Does this work for you?  
A github action running the tests could help with that. Happy to create a PR if this is wanted.

fiatjaf:
It works for me. There is a GitHub action running the tests.

It fails for you because your NodeJS doesn't have WebCrypto stuff. Newer versions should have that.

PR #90, opened 2023-01-16
docs: add node version info

escapedcat:
Might help future users with this: https://github.com/nbd-wtf/nostr-tools/issues/88#issuecomment-1384240793


fiatjaf:
Sorry, but this is unnecessary.

PR #94, opened 2023-01-21
incoming message queue, alreadyHaveEvent check, msg.send catch

mmalmi:
* Add incoming messages to processing queue and process them on setInterval, so we don't block UI processing
* Optional `alreadyHaveEvent` param to `relayInit` so you can skip JSON.parse and event callback, if you already have the event in memory
* Catch errors in `ws.send(msg)`
* Resolve `untilOpen` on `ws.open`

### DIFF
--- a/relay.ts
+++ b/relay.ts
@@ -31,10 +31,11 @@ type SubscriptionOptions = {
   id?: string
 }
 
-export function relayInit(url: string): Relay {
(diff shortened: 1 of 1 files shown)


fiatjaf:
This is so weird, I was looking into adding this exactly this moment, after you told me these tricks more than a week ago.

PR #96, opened 2023-01-31
Fullname should be provided to nip05's queryProfile

ktecho:
When using queryProfile from nip05, you should provide a fullname, not just the @domain part, so I'm "fixing" the example here.


fiatjaf:
No, if you pass just a domain it will automatically query for `_@domain`, which is the expected behavior according to NIP-05.

PR #97, opened 2023-01-31
nip05 supports uppercase

ktecho:
nip05 says `NIP-05 assumes the <local-part> part will be restricted to the characters a-z0-9-_., case insensitive`

So a lot of people is starting the names with uppercase. See here:

https://nostr-check.com/.well-known/nostr.json

So I think we should change the regex to accept lowercase or uppercase.

Another way to do it would be to do a `.toLowerCase` at the beginning, but then we would need to do this search ignoring the case:

```
if (!res?.names?.[name])
```

So maybe for now this is enough?

### DIFF
--- a/nip05.ts
+++ b/nip05.ts
@@ -36,7 +36,7 @@ export async function queryProfile(
     name = '_'
   }
 
-  if (!name.match(/^[a-z0-9-_]+$/)) return null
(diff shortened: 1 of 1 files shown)


fiatjaf:
Indeed, thank you.

PR #103, opened 2023-02-07
Path API

mmalmi:
Path API for Nostr, built on [NIP33](https://github.com/nostr-protocol/nips/blob/master/33.md) replaceable-by-tag events of kind 30000.
```
const path = new Path(myPublishFn, mySubscribeFn, myUnsubscribeFn, { authors: [myPubKey} })
path.set('reactions/[noteID]', '😎')
path.get('reactions/[noteID]', (value, path, event) => console.log(event.pubkey, 'reacted with', value))
```

In-memory caches the most recent event for the subscribed path, and only calls back with the most recent value.

This API allows us to build all kinds of applications on top of Nostr (github replacement for example) without having to
specify new event kinds all the time and implement them in all applications and relays.

In [iris](https://github.com/irislib/iris-messenger) I'm currently using this to save `notifications/lastOpened`, `chats/[pubkey]/lastOpened` and `settings/colorScheme`.

I reckon having to specify publish, subscribe, unsubscribe etc. in the constructor can be a bit cumbersome. If you deem better, I can publish this API as a part of iris-lib which would also automatically set those functions.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -2,6 +2,7 @@ export * from './keys'
 export * from './relay'
 export * from './event'
 export * from './filter'
+export * from './path'
 
 export * as nip04 from './nip04'
 export * as nip05 from './nip05'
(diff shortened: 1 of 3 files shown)


fiatjaf:
Sorry, I forgot to answer this at the time. I still don't know how to answer because the use case is not clear in my mind. It's been so long I think I'll close it and wait.

PR #105, opened 2023-02-09
fix(pool): clear timeout on cancelation signal

fernandolguevara:


fiatjaf:
This is clearing the timeout on the timeout callback itself. Is this necessary?

fernandolguevara:
you're right

PR #106, opened 2023-02-09
chore(dx): add common scripts

fernandolguevara:


fernandolguevara:
mortals aren't used to `just` 🤷‍♂️

fiatjaf:
I understand, but I am not used to `package.json` scripts.

fernandolguevara:
em

`https://github.com/fiatjaf/nos2x/blob/master/package.json#L20`

PR #110, opened 2023-02-12
Add field `aliases` to `ProfilePointer` for nip05

dallaslu:
In NIP-05, there may be multiple names corresponding to the same Pubkey in `nostr.json`. For example:
```json
{
  "names": {
    "jb55": "32e1827635450ebb3c5a7d12c1f8e7b2b514439ac10a67eef3d9fd9c5c68e245",
    "minebot": "6cad545430904b84a8101c5783b65043f19ae29d2da1076b8fc3e64892736f03",
    "_": "32e1827635450ebb3c5a7d12c1f8e7b2b514439ac10a67eef3d9fd9c5c68e245"
  }
}
```
This may be for compatibility purposes. And In some client implementations, only the domain name part is displayed for nip05, which has some risks. Once the attacker has obtained the identity of nip05 (or a privatekey, such as `minebot` in this case) under the same domain name, he can modify his profile to disguise as anyone with the same domain name.

So in the future, the client may have to highlight when the name of the profile is inconsistent with the name of nip05. But the real owner, if you use `_@domain` will be treated wrongly. Therefore, if the alias is obtained at the same time of verification, it will help to handle it correctly. For example:

```javascript
nip05.queryProfile('jb55.com');
```
and
```javascript
nip05.queryProfile('jb55@jb55.com');
```

Both will get:
```javascript
{
  pubkey: '32e1827635450ebb3c5a7d12c1f8e7b2b514439ac10a67eef3d9fd9c5c68e245',
  aliases: ['jb55', '_']
}
```

### DIFF
--- a/nip05.ts
+++ b/nip05.ts
@@ -46,9 +46,11 @@ export async function queryProfile(
 
   let pubkey = res.names[name] as string
   let relays = (res.relays?.[pubkey] || []) as string[] // nip35
+  let aliases = Object.keys(res.names).filter(k => res.names[k] === pubkey)
 
   return {
     pubkey,
-    relays
+    relays,
+    aliases
   }
 }
--- a/nip19.ts
+++ b/nip19.ts
@@ -8,6 +8,7 @@ const Bech32MaxSize = 5000
 export type ProfilePointer = {
   pubkey: string // hex
   relays?: string[]
+  aliases?: string[]
 }
 
 export type EventPointer = {


fiatjaf:
Implementations shouldn't look at aliases, they should only try `response["names"][nameTheyAreLookingFor]`. Dynamic servers will return an object with a single name, which is the name specified in the name `?name=`.

If servers return a static file with multiple names in it that fine, but we shouldn't rely on that or expect that or treat that as a feature.

fiatjaf:
I don't understand the vulnerability you found, or attack vector, or inconsistency. What is the inconsistency? If you're looking for `_` you get `32e1827635450ebb3c5a7d12c1f8e7b2b514439ac10a67eef3d9fd9c5c68e245`, if you're looking for `jb55` you get `32e1827635450ebb3c5a7d12c1f8e7b2b514439ac10a67eef3d9fd9c5c68e245`, it's all good. No?

dallaslu:
Take the Damus app as an example. 

There is a target user, sign a kind-0 event with content:

```javascript
{
  banner: "https://alice.com/b.jpg",
  website: "https://alice.com/",
  nip05: "alice@getalby.com",
  picture: "https://alice/p.png",
  lud16: "alice@getalby.com",
  display_name: "Alice",
  about: "Blabla",
  name: "alice"
}
```

When people view their profile page, they will see the profie name and nip05 id displayed as:

> @alice [checkmark] getalby.com

Now there is an attacker, sign a kind-0 event with content:

```javascript
{
  banner: "https://alice.com/b.jpg",
  website: "https://alice.com/",
  nip05: "fakealice@getalby.com",
  picture: "https://alice/p.png",
  lud16: "fakealice@getalby.com", // or alice@getalby.com
  display_name: "Alice",
  about: "Blabla",
  name: "alice"
}
```

Will be displayed as the same:

> @alice [checkmark] getalby.com

Their profile pages may look exactly the same, because followers, prefix or suffix (may be eight letters) of npub, notes, and so on can be deliberately forged. For a person who has not yet followed the target user, it is difficult to distinguish.

However, due to design reasons, npubs are not fully displayed by default in the app. To solve this problem, we can only strengthen the display of nip05 information. 

For some static websites with few users, they may list all the names in nostr.json, such as  <https://jb55.com/.well-known/nostr.json>. And use `_@domain.tld` or `domain.tld` as nip05 identity.

```javascript
{
  nip05: "_@jb55.com",
  name: "jb55"
}
```

This will make it difficult for app to compare profile name and nip05 name. So we need to know its alias that mapped to the same pubkey, if provided.

fiatjaf:
The name and the nip05 identifiers are completely independent. I don't see what we gain by mixing both. It will just create confusion (at least I am very confused by what you're saying).

If Damus is displaying a "verified" icon and not the full nip05 address and not telling what the hell is "verified" then this should be a bug report on Damus, not a spec change.

fiatjaf:
Please reply if you disagree. I'm closing just for my own sanity.

PR #111, opened 2023-02-12
Browser tests

mistermoe:
Wanted to start by giving props to the maintainers of this lib. it's easy/straightforward to use and the API surface is pretty damn intuitive. 

Wondering if there's any interest in adding automated browser tests? I figured it may be useful from an integrity perspective because the library is intended to be usable in both the browser and node runtimes (potentially react native as well?). The hope is that it'll be easier to catch potential breaking changes. 

I went ahead and just did it in this PR thinking that it might be more useful _see_/_use_ it vs. just read about it in a github issue. 

image

PR includes the ability to run the existing test suite in chrome, firefox, and webkit. There's definitely several different approaches to doing this. The approach used in this PR was driven by the goal of changing little to nothing about the existing test suite. 

Additionally, I made a few minor changes to reduce the steps required to use this library in node. Tried my best to isolate each change in separate commits so that it's easier to cherry-pick specific changes and ditch others if needed.

Lmk what y'all think


README.md
```diff
@@ -177,12 +171,6 @@ console.log(profile.relays)
 // prints: [wss://relay.damus.io]
 ```
 
-To use this on Node.js you first must install `node-fetch@2` and call something like this:
-
-```js
-nip05.useFetchImplementation(require('node-fetch'))
```
fiatjaf, line 183: Why did you remove this?

nip05.ts
```diff
@@ -1,22 +1,21 @@
+import crossFetch from 'cross-fetch'
```
fiatjaf, line 1: Isn't this going to include this dependency for everybody regardless of the environment they are?

nip04.test.js
```diff
@@ -1,6 +1,5 @@
 /* eslint-env jest */
 
-globalThis.crypto = require('crypto')
```
fiatjaf, line 3: Why?

mistermoe: reasons were:
* it's not being used anywhere in the file
* node runtime tests pass without it
* browser tests fail with it
* `globalThis.crypto` isn't being used by any dependencies because all `@noble` and `@scure` libs are set up to use a runtime-appropriate `crypto` 

e.g. `noble-secp256k1` requires `crypto` [here](https://github.com/paulmillr/noble-secp256k1/blob/main/index.ts#L4-L6), but this line is never loaded by `esbuild` when bundling for the browser because `crypto` is set to `false` [here](https://github.com/paulmillr/noble-secp256k1/blob/main/package.json#L27-L29).  Then, `@noble` and `@scure` libs that use crypto also have conditionals in place to use the appropriate crypto lib whenever needed. e.g. [assigning crypto](https://github.com/paulmillr/noble-secp256k1/blob/main/index.ts#L1533-L1536), [using appropriate crypto](https://github.com/paulmillr/noble-secp256k1/blob/main/index.ts#L1583-L1591)

relevant esbuild docs:
>  If a package specifies a map for the [browser](https://gist.github.com/defunctzombie/4339901/49493836fb873ddaa4b8a7aa0ef2352119f69211) field in its package.json file, esbuild will use that map to replace specific files or modules with their browser-friendly versions. For example, a package might contain a substitution of [path](https://nodejs.org/api/path.html) with [path-browserify](https://www.npmjs.com/package/path-browserify).

relevant [browser property spec](https://github.com/defunctzombie/package-browser-field-spec#ignore-a-module)
> You can simply prevent a module or file from being loaded into a bundle by specifying a value of false for any of the keys. This is useful if you know certain codepaths will not be executed client side but find it awkward to split up or change the code structure.

Failing browser tests with `globalThis.crypto = require('crypto')`:

```
WebKit 605.1.15 (Mac OS 10.15.7) ERROR
  TypeError: Attempted to assign to readonly property.
  @browser-tests/nip04.test.js:51255:13
  global code@browser-tests/nip04.test.js:51266:3
```

```
Firefox 109.0 (Mac OS 10.15) ERROR
  TypeError: setting getter-only property "crypto"
  NostrTools<@browser-tests/nip04.test.js:51255:23
  @browser-tests/nip04.test.js:51266:3
```

```
Chrome Headless 109.0.5414.119 (Mac OS 10.15.7) ERROR
  Uncaught TypeError: Cannot set property crypto of #<Window> which has only a getter
  TypeError: Cannot set property crypto of #<Window> which has only a getter
      at browser-tests/nip04.test.js:51255:21
      at browser-tests/nip04.test.js:51266:3
```

but maybe i'm missing why that line is there in the first place?

README.md
mistermoe: adding it back. will resurface later in a separate PR. feels out of scope here

nip05.ts
mistermoe: yeah you're right. reverting this back to how it was bc feels out of scope here. will resurface in a separate PR later

README.md
mistermoe: [done](https://github.com/nbd-wtf/nostr-tools/compare/4fd83691968a7dacbc179fd08a571b4865329493..829e24b5bc65725b76172d764316e146e4ebe572)

using `cross-fetch` in test instead of `node-fetch` to support running same test in browser and node. [relevant commit](https://github.com/nbd-wtf/nostr-tools/pull/111/commits/0d577c67fd414ef26ea546efe7984a2e9849b5cb)

nip05.ts
mistermoe: [done](https://github.com/nbd-wtf/nostr-tools/compare/4fd83691968a7dacbc179fd08a571b4865329493..829e24b5bc65725b76172d764316e146e4ebe572)

using `cross-fetch` in test instead of `node-fetch` to support running same test in browser and node. [relevant commit](https://github.com/nbd-wtf/nostr-tools/pull/111/commits/0d577c67fd414ef26ea546efe7984a2e9849b5cb)

nip04.test.js
fiatjaf: It is there because some people were complaining it didn't run on their Node.js -- although it does run on mine without it. These people had the same Node.js version I did so that was confusing.

fiatjaf: Bundling `nostr-tools` for the browser does work today, though. But I see that if you bundle the tests they will include this and everything will break.

I guess we can just remove this line and ask people to do it themselves if they want? Or prepend that to the file before running tests in Node.

mistermoe: > It is there because some people were complaining it didn't run on their Node.js -- although it does run on mine without it. These people had the same Node.js version I did so that was confusing.

ok cool thx for the context. i'm using node `v19.2.0`. i switched to `lts/hydrogen (v18.12.1)` and ran the tests and `nip04.test.js` failed without the `require`. worked with it.  then went further back to `lts/gallium (v16.19.0)`.  failed with _and_ without the `require`. 

`v16.19.0` fails because `crypto` doesn't have `subtle` as a top-level property. it's [`crypto.webcrypto.subtle`](https://nodejs.org/docs/latest-v16.x/api/webcrypto.html#web-crypto-api). this means that `nip04`'s `encrypt` and `decrypt` will bork for anyone using `nostr-tools` on node v16.x

`v18.12.1` and `v19.2.0` failed without the `require` because of how the `cjs` is bundled. 

> I guess we can just remove this line and ask people to do it themselves if they want? Or prepend that to the file before running tests in Node.

na just fixed the `cjs` bundle, removed the `require` from `nip04.test.js`, and added this to the top of `nip04.js`:

```
import nodeCrypto from 'crypto'
let crypto = nodeCrypto?.webcrypto ?? globalThis.crypto
```

this makes it so `encrypt` and `decrypt` work in the browser and node v16.x, v18.x, and v19x. just need remove some debug logs i added and i'll push up the fix

mistermoe: pushed. older versions of node are supported now. specifically, 16.x 18.x and 19.x

fiatjaf:
I'm sorry about this PR. I wanted to merge it at some point, but I don't remember why I didn't. Now it's definitely too late.

PR #122, opened 2023-02-18
allow alreadyHaveEvent fn to return boolean | Promise<boolean>

BilligsterUser:


fiatjaf:
I have the feeling that this kills all the performance gains from `alreadyHaveEvent`.

fiatjaf:
But it was just a feeling, I don't really know!

fiatjaf:
I see, then we must support it. Can you profile a similar case on https://jsbench.me/? Just to see if it's worth using a different way for async people or if just making the current way async will be good enough.

PR #125, opened 2023-02-19
Make `esm` bundle usable in node 

### LINKED ISSUE #124: using esm imports doesn't work in node
README shows example of esm style imports

image

running ^ causes the following:
```
import { generatePrivateKey } from 'nostr-tools';
         ^^^^^
SyntaxError: Named export 'generatePrivateKey' not found. The requested module 'nostr-tools' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'nostr-tools';
const { generatePrivateKey } = pkg;
```

using node `v19.2.0` and npm `v8.19.3`

mistermoe:
esm imports don't work currently because node decides whether a package is cjs or esm based on the `type` property in the nearest `package.json`. default value of `type` is `commonjs` which causes import/require import whatever is set for `main` which happens to be the cjs bundle. 

This PR fixes #124  by:
* auto-generating a minimal `package.json in `lib` dir nearest the esm bundle
  * Noticed this package uses `cjs` bundle in all tests so adding `type: "module"` to root package.json would cause things to break.
* including the `exports` property in the root package.json to allow node environments to use cjs or esm bundles

### DIFF
--- a/build.js
+++ b/build.js
@@ -1,5 +1,6 @@
 #!/usr/bin/env node
 
+const fs = require('fs')
 const esbuild = require('esbuild')
 
 let common = {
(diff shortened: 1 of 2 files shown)


alexgleason:
Can we get a release for this please?

PR #140, opened 2023-02-27
add type definition

### LINKED ISSUE #138: Deno import doesn't work properly
Something is wrong in Deno when importing with `npm:nostr-tools`. I hoped #125 would fix it, but no dice.

The functions do not exist. `import * as nostr from 'npm:nostr-tools'` seems to work okay, but I don't get types if I do that.

I think something is wrong with the export.

image

Importing from jspm _does_ work, but causes problems with the lockfile: https://dev.jspm.io/nostr-tools@1.6.0

image

Most npm modules work on Deno, so I think something is wrong here, and that the fix would likely improve things in a web environment.

BilligsterUser:
fixes #138

### DIFF
--- a/justfile
+++ b/justfile
@@ -12,5 +12,9 @@ test: build
 testOnly file: build
     jest {{file}}
 
+types: 
+    npx tsc
+    
 publish: build
+    types
     npm publish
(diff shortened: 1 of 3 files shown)


alexgleason:
Thank you!

PR #148, opened 2023-03-04
close relay's websocket only if it's in state OPEN

ramigs:
### Motivation

test `pool.test.js` was failing with error `TypeError: Cannot read properties of undefined (reading 'sendCloseFrame')`.
seems that (at least one) of the relays in the `relays` array is forcefully closing the connection, which was causing the error mentioned above, when trying to close the pool in `afterAll`.

`jest` was not exiting due to some websockets left unclosed. I think this is the reason that GitHub Actions workflow `test` has been failing recently (by exceeding the maximum execution time).

jest-fail

### DIFF
--- a/relay.ts
+++ b/relay.ts
@@ -310,8 +310,9 @@ export function relayInit(
       listeners = {connect: [], disconnect: [], error: [], notice: []}
       subListeners = {}
       pubListeners = {}
-
-      ws?.close()
+      if (ws.readyState === WebSocket.OPEN) {
+        ws?.close()
+      }
     },
     get status() {
       return ws?.readyState ?? 3


fiatjaf:
It's still failing.

ramigs:
fail seems to be unrelated to this change.
it failed in the PR's pipeline (in `relay.test.js` with `"Exceeded timeout of 5000 ms for a testfalse.`, seems relay in the test was not responding when test executed)
https://github.com/nbd-wtf/nostr-tools/actions/runs/4331616113/jobs/7563773501

but all tests executed successfully in the pipeline execution for `master`
https://github.com/nbd-wtf/nostr-tools/actions/runs/4331752560
ran tests locally multiple times and always succeeding.

maybe in the future tests will need to be refactored to take into account eventual unavailability of relays.

fiatjaf:
Wow, yes, indeed, they are succeeding on `master`. Amazing, thank you!

Issue #157, opened 2023-03-12
nip-01: are prefixed authors and ids supported?


offbyn:
> The ids and authors lists contain lowercase hexadecimal strings, which may either be an exact 64-character match, or a prefix of the event value.

https://github.com/nostr-protocol/nips/blob/master/01.md

I am trying to subscribe with a filter for ids of ['0000'] with the intention to only receive events with some minimum POW eventId.

After figuring out the events are received but 'event' is not fired (see https://github.com/scsibug/nostr-rs-relay/issues/104) I wonder if prefixed id's are supported by `nostr-tools` or if this is a bug.

offbyn:
in this fiddle events starting with 0000 are received but somehow nostr-tools doesn't fire the event https://jsfiddle.net/syr4k235/

fiatjaf:
They were technically supported but as you found out there was a bug, then the support was officially removed from the NIPs.

PR #158, opened 2023-03-12
nip01: add support for filter prefix in authors and ids

offbyn:
so clients can filter events by prefix ids as described in nip-01, i.e. to subscribe to mined events starting with zeroes or to add some privacy for clients that may not want to disclose the exact filter.

this is only a POC and does not support non-aligned (odd-length) hex prefixes.

see also https://github.com/scsibug/nostr-rs-relay/issues/104

part of https://github.com/nbd-wtf/nostr-tools/issues/157

### DIFF
--- a/filter.test.js
+++ b/filter.test.js
@@ -37,6 +37,16 @@ describe('Filter', () => {
       expect(result).toEqual(false)
     })
 
+    it('should return true when the event id starts with a prefix', () => {
+      const filter = {ids: ['22', '00']}
+
+      const event = {id: '001'}
+
+      const result = matchFilter(filter, event)
+
+      expect(result).toEqual(true)
+    })
+
     it('should return false when the event kind is not in the filter', () => {
       const filter = {kinds: [1, 2, 3]}
 
@@ -126,6 +136,20 @@ describe('Filter', () => {
       ]
 
       const event = {id: '789', kind: 3, pubkey: 'ghi'}
+
+      const result = matchFilters(filters, event)
+
+      expect(result).toEqual(true)
+    })
+
+    it('should return true when at least one prefix matches the event', () => {
+      const filters = [
+        {ids: ['1'], kinds: [1], authors: ['a']},
+        {ids: ['4'], kinds: [2], authors: ['d']},
+        {ids: ['9'], kinds: [3], authors: ['g']}
+      ]
+
+      const event = {id: '987', kind: 3, pubkey: 'ghi'}
 
       const result = matchFilters(filters, event)
 
--- a/filter.ts
+++ b/filter.ts
@@ -15,10 +15,17 @@ export function matchFilter(
   filter: Filter,
   event: Event
 ): boolean {
-  if (filter.ids && filter.ids.indexOf(event.id) === -1) return false
+  if (filter.ids && filter.ids.indexOf(event.id) === -1) {
+    if (!filter.ids.some(prefix => event.id.startsWith(prefix))) {
+      return false
+    }
+  }
   if (filter.kinds && filter.kinds.indexOf(event.kind) === -1) return false
-  if (filter.authors && filter.authors.indexOf(event.pubkey) === -1)
-    return false
+  if (filter.authors && filter.authors.indexOf(event.pubkey) === -1) {
+    if (!filter.authors.some(prefix => event.pubkey.startsWith(prefix))) {
+      return false
+    }
+  }
 
   for (let f in filter) {
     if (f[0] === '#') {


offbyn:
@fiatjaf would you accept such an approach / cACK?

if yes, then I would look into support for non-aligned (odd-length) hex prefixes and get this PR ready.

PTAL

fiatjaf:
I don't understand what is the benefit of adding this to the library.

offbyn:
the user story is: as a client I want to have the option to filter the feed and only show events that were mined with nip-13 pow.

the problem: this is already possible by checking every received event if the event-id starts with certain amount of 0's (zeros).  but each event still has to be downloaded (impact on bandwidth) and verified (impact on performance). ideally the client would tell a relay: "hey please only send me events starting with `00000`" 

the benefit is: relay has to send less data and client saves bandwidth and cpu. additionally this is already specified in nip-01 and at least nostr-rs-relay already supports it.

but it seems that nostr-tools checks for exact matches and does not support prefix id's, therefore the events are received but are not emitted by nostr-tools library.

(sorry I am slow, I mostly work on the weekends on my nostr project)

offbyn:
please also see this comment with example jsfiddle https://github.com/nbd-wtf/nostr-tools/issues/157#issuecomment-1465210163

fiatjaf:
Right, but the library currently already supports this, there is nothing to be changed. What am I missing?

offbyn:
> Right, but the library currently already supports this, there is nothing to be changed. What am I missing?

ah then it might be a bug, the following code should log an event startign with event-id 0000, such as https://www.nostr.guru/e/0000c975913d07074ce3a83fe579da71f49dca2c3aed2a31a1b8fd754d0585cc

```js
 const subscription = relay.sub([
    {
      ids: ["0000"], // only subscribe to id's with prefix 0
      kinds: [1],
      limit: 2
    }
  ]);
  subscription.on("event", (event) => {
    console.log("event", relay.url, event);
  });
```

https://jsfiddle.net/dahcjyu4/

offbyn:
here it looks for exact matches (with indexOf) https://github.com/nbd-wtf/nostr-tools/blob/ce081bb4cb7ce5deef7f637e403b68509eae01d6/filter.ts#L18

>  but the library currently already supports this

I can't see that in the code, where should it allow prefix ids?

fiatjaf:
I see it now. Sorry for my obtuseness.

fiatjaf:
Can you fix the conflict?

filter.ts
```diff
@@ -15,7 +15,11 @@ export function matchFilter(
...
   event: Event & {id: string}
 ): boolean {
-  if (filter.ids && filter.ids.indexOf(event.id) === -1) return false
+  if (filter.ids && filter.ids.indexOf(event.id) === -1) {
+    if (!filter.ids.some(prefix => prefix.length < 64 && event.id.startsWith(prefix))) {
+      return false
+    }
+  }
   if (filter.kinds && filter.kinds.indexOf(event.kind) === -1) return false
   if (filter.authors && filter.authors.indexOf(event.pubkey) === -1)
```
fiatjaf, line 24: shouldn't we also allow prefixes here?

filter.ts
```diff
@@ -15,7 +15,11 @@ export function matchFilter(
   filter: Filter,
   event: Event & {id: string}
 ): boolean {
-  if (filter.ids && filter.ids.indexOf(event.id) === -1) return false
+  if (filter.ids && filter.ids.indexOf(event.id) === -1) {
+    if (!filter.ids.some(prefix => prefix.length < 64 && event.id.startsWith(prefix))) {
```
fiatjaf, line 19: remove this `prefix.length` check, it's unnecessary

filter.ts
offbyn: I agree, ids and authors https://github.com/nostr-protocol/nips/blob/master/01.md

filter.ts
offbyn: I'll remove, I just wanted to make it explicit that prefix is only if it's smaller than 64.

> The ids and authors lists contain lowercase hexadecimal strings, which may either be an exact 64-character match, or a prefix of the event value.

https://github.com/nostr-protocol/nips/blob/master/01.md

filter.test.js
```diff
@@ -9,6 +9,19 @@ test('test if filters match', () => {
...
+      bad: [{id: '001'}]
+    },
+    {
+      filters: [{ids: ['0000', '000111']}],
+      good: [
+        {id: '000011'},
+        {id: '000111'}
+      ],
+      bad: [{id: '001111'}]
+    },
```
offbyn, line 24: does that make sense?

filter.ts
offbyn: maybe it would also make sense in "#e" and "#p", but that is not in nip-01 so out of scope for this PR.

filter.ts
offbyn: removed

filter.test.js
```diff
@@ -131,6 +141,20 @@ describe('Filter', () => {
...
+        {ids: ['4'], kinds: [2], authors: ['d']},
+        {ids: ['9'], kinds: [3], authors: ['g']}
+      ]
+
+      const event = {id: '987', kind: 3, pubkey: 'ghi'}
+
+      const result = matchFilters(filters, event)
+
+      expect(result).toEqual(true)
+    })
```
offbyn, line 157: let me know if this test makes sense or if you'd like other tests.

offbyn:
@fiatjaf rebased, updated tests, added some comment and added support for authors.

PTAL

offbyn:
actually I noticed prettier would format the filter arguments on one line, but it was already like this.. so I'll keep as it was.

```diff
diff --git a/filter.ts b/filter.ts
index 263b853..5085256 100644
--- a/filter.ts
+++ b/filter.ts
@@ -11,10 +11,7 @@ export type Filter = {
   [key: `#${string}`]: string[]
 }
 
-export function matchFilter(
-  filter: Filter,
-  event: Event
-): boolean {
+export function matchFilter(filter: Filter, event: Event): boolean {
   if (filter.ids && filter.ids.indexOf(event.id) === -1) {
     if (!filter.ids.some(prefix => event.id.startsWith(prefix))) {
       return false
@@ -47,10 +44,7 @@ export function matchFilter(
   return true
 }
 
-export function matchFilters(
-  filters: Filter[],
-  event: Event
-): boolean {
+export function matchFilters(filters: Filter[], event: Event): boolean {
   for (let i = 0; i < filters.length; i++) {
     if (matchFilter(filters[i], event)) return true
   }
```

filter.test.js
fiatjaf: I think most relays expect pairs of hex characters and will break if you pass an odd number.

fiatjaf: But I'm fine with this for the tests, it does the job, we can fix it later.

PR #162, opened 2023-03-18
Fix nip05 README example

ebrakke:
`queryProfile` expects a full address with `@`


ebrakke:
Nvm, I see that this is actually valid. I wasn't aware that if you just pass in a domain, the implied name would be `_`

PR #164, opened 2023-03-23
add 17 test cases for event.js

sepehr-safari:
### Added 17 test cases for event.js described as below:

**getBlankEvent** should return a blank event object.

**finishEvent** should create a signed event from a template.

**serializeEvent** should serialize a valid event object.
**serializeEvent**  should throw an error for an invalid event object.

**getEventHash** should return the correct event hash.

**validateEvent** should return true for a valid event object.
**validateEvent** should return false for a non object event.
**validateEvent** should return false for an event object with missing properties.
**validateEvent** should return false for an empty object.
**validateEvent** should return false for an object with invalid properties.
**validateEvent** should return false for an object with an invalid public key.
**validateEvent** should return false for an object with invalid tags.

**verifySignature** should return true for a valid event signature.
**verifySignature** should return false for an invalid event signature.
**verifySignature** should return false when verifying an event with a different private key.

**signEvent** should sign an event object.
**signEvent** should not sign an event with different private key.

### DIFF
--- a/event.test.js
+++ b/event.test.js
@@ -307,5 +307,34 @@ describe('Event', () => {
       expect(sig.length).toEqual(128)
       expect(isValid).toEqual(true)
     })
+
+    it('should not sign an event with different private key', () => {
+      const privateKey =
+        'd217c1ff2f8a65c3e3a1740db3b9f58b8c848bb45e26d00ed4714e4a0f4ceecf'
+      const publicKey = getPublicKey(privateKey)
+
+      const wrongPrivateKey =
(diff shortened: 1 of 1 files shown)


fiatjaf:
And these are passing?

sepehr-safari:
> And these are passing?

Absolutely, Yeap.

sepehr-safari:
Thanks for your attention @fiatjaf 🤙
I'm writing a bunch of test cases for other files, as well.
Will submit further PRs.

PR #168, opened 2023-03-27
validateEvent: use assertion function

alexgleason:
Improves the return type of validateEvent so it can act as a guard. This allows passing `unknown` into it to get back an `UnsignedEvent`. If you pass a more fully-qualified type like `Event`, it'll return the more detailed type. This is a backwards-compatible change.


offbyn:
related `is` doc https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

fiatjaf:
This is great, I would never know.

PR #172, opened 2023-04-02
 Fix explicit NIP-10 root/reply/mention markers parsed incorrectly 

futpib:
Also fixed pubkey relays being ignored.
Also fixed code style with prettier.
Sorry for the original #171 PR being a bit of a mess.


fiatjaf:
Should we rename that array from `pubkeys` to `profiles`?

futpib:
#173

PR #177, opened 2023-04-06
fix: mnemonic strength should be 256 bits.

susumuota:
`bip39.generateMnemonic(wordlist)` generates 128 bit strength key without `strength` option.
In my opinion, it should be 256 bit.

* https://github.com/paulmillr/scure-bip39/blob/dfadcd338cb0954e60627fccaa658671e5d33257/src/index.ts#L39
* https://github.com/v0l/snort/pull/435

### DIFF
--- a/nip06.ts
+++ b/nip06.ts
@@ -18,7 +18,7 @@ export function privateKeyFromSeedWords(
 }
 
 export function generateSeedWords(): string {
-  return generateMnemonic(wordlist)
+  return generateMnemonic(wordlist, 256)
 }
 
 export function validateWords(words: string): boolean {


fiatjaf:
Is that 24 words?

susumuota:
Yes it will generate 24 words mnemonic phrase.

When the user creates a private key with `generatePrivateKey()`, it creates a key with 256 bit strength.
On the other hand, if they create a key with `generateSeedWords`, it creates a key with 128 bit strength.

This sounds inconsistent to me. I cannot find a strong reason to keep it 128 bit instead of this inconsistency.

susumuota:
I have checked [BIP32](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki), [BIP39](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki), etc. Please correct me if I am misunderstanding something.

[NIP-06](https://github.com/nostr-protocol/nips/blob/master/06.md) derives private key from master seed (128-256 bit) with fixed path `m/44'/1237'/0'/0/0` (assume `account` is `0` here).
So the NIP-06 private key corresponds to `m/0` in [the image below](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#master-key-generation) (`depth = 1` and `index = 0`).

image

This means that the NIP-06 private key is generated deterministically from the master seed because the `depth` and `index` are fixed.

In this case, the strength of the NIP-06 private key is equal to the strength of the master seed. If the master seed is 128-bit, malicious attackers only need to perform a brute-force attack on the 128-bit key space because they don't need to try various `depth` and `index`.

On the other hand, in the regular use case of BIP32, attackers need to spend computational resources to try various `depth` and `index` to perform brute-force attacks. Thus, the whole process might be harder than a simple one-time 128-bit brute-force attack.

Also, 256-bit strength master seed is `advised` by [BIP32](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#master-key-generation).

> Generate a seed byte sequence S of a chosen length (between 128 and 512 bits; 256 bits is advised) from a (P)RNG.

In my opinion, the master seed strength must be 256 bits. Or, if you want to keep the default at 128 bits, it's better to inform users widely:

```
Your private key is 2^128 times weaker than other 256-bit keys.
```

I think the difference in usability between 12 or 24 words is a minor issue compared to the strength of the private key.

fiatjaf:
Sorry for not answering. I think no one should be using NIP-06 anyway, so I think I won't make any changes. Eventually I'll remove NIP-06 this from this library.

PR #178, opened 2023-04-07
scripts: add npm build, format and test scripts

offbyn:
in addition to the just tasks, this commit adds npm scripts back, for convenience.
    
example taks:
- npm test
- npm test filter.test.js
- npm run build
- npm run format
    
in yarn it should just work without 'run' i.e. 'yarn build'.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,11 @@
     "client",
     "nostr"
   ],
+  "scripts": {
+    "build": "node build",
+    "format": "prettier --plugin-search-dir . --write .",
+    "test": "node build && jest"
+  },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


offbyn:
& this would help me personally next time when I want to contribute, so that I don't have to search the project for how the tooling works again.

also so that people don't have to install tools globally.

fiatjaf:
The biggest problem with that is that you must call `npm run` from inside the other `npm run`, which means it doesn't work if you only have `yarn`.

fiatjaf:
Yes.

offbyn:
PTAL update PR kept build, format and test.

didn't test on yarn, but I think it should work.

PR #188, opened 2023-04-18
nip19: improve return type

alexgleason:
Makes the return type of `nip19.decode` a [discriminated union](https://knowledgehub.hashnode.dev/typescript-discriminated-unions-a-guide-to-understanding), allowing you to use a `switch` statement on `result.type` and then `result.data` will be automatically inferred.

Eg:

```ts
const result = decode('1234...')

switch (result.type) {
  case 'nprofile':
    result.data // ProfilePointer inferred
    break
  case 'npub':
    result.data // string inferred
    break
}
```

### DIFF
--- a/nip19.ts
+++ b/nip19.ts
@@ -23,10 +23,15 @@ export type AddressPointer = {
   relays?: string[]
 }
 
-export function decode(nip19: string): {
-  type: string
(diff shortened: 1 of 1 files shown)


fiatjaf:
TypeScript is actually pretty nice despite being JavaScript.

PR #195, opened 2023-04-23
Rename signEvent to getSignature

alexgleason:
The existing `signEvent` function is named incorrectly. It should be called `getSignature`, which matches its partner function `getEventHash`. This is because it returns a _string_ instead of a signed event. Even NIP-07's `window.nostr.signEvent` function is correct, returning the signed event, but the one in this repo isn't.

This is a breaking change, but I think it's needed. The name of this function interferes with actual `signEvent` implementations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ let pk = getPublicKey(sk) // `pk` is a hex string
 import {
   validateEvent,
   verifySignature,
-  signEvent,
+  getSignature,
   getEventHash,
   getPublicKey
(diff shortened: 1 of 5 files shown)


fiatjaf:
I agree with you, but I'm afraid this will break too much. Maybe we should keep both for a while and change the docs and add a deprecation notice or something, then remove it after 6 months?

alexgleason:
Good call. I updated the branch and restored the `signEvent` function with a deprecation warning.

PR #204, opened 2023-05-06
Improve event types

alexgleason:
Fixes typescript issues when trying to pass events of unknown kind into functions that shouldn't care. Functions that don't care about kind should accept an `Event<number>`.

More comments below.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -103,7 +103,6 @@
     "no-octal-escape": 2,
     "no-path-concat": 0,
     "no-proto": 2,
-    "no-redeclare": 2,
     "no-regex-spaces": 2,
     "no-return-assign": 0,
     "no-self-assign": 2,
--- a/event.ts
+++ b/event.ts
@@ -6,7 +6,6 @@ import {getPublicKey} from './keys'
 
 /* eslint-disable no-unused-vars */
 export enum Kind {
-  UndefinedKindNumber = 255,
   Metadata = 0,
   Text = 1,
   RecommendRelay = 2,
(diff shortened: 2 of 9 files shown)


.eslintrc.json
```diff
@@ -103,7 +103,6 @@
     "no-octal-escape": 2,
     "no-path-concat": 0,
     "no-proto": 2,
-    "no-redeclare": 2,
```
alexgleason, line 106: I had to remove this to make TypeScript function overloading work. We could enable the TypeScript version if we want, but I didn't see the point. I think TS will warn you anyway.

event.ts
```diff
@@ -20,6 +19,7 @@ export enum Kind {
   ChannelMessage = 42,
   ChannelHideMessage = 43,
   ChannelMuteUser = 44,
+  Blank = 255,
```
alexgleason, line 22: I changed the name of this from #200. The other name was overly verbose.

nip57.ts
```diff
@@ -13,7 +13,7 @@ export function useFetchImplementation(fetchImplementation: any) {
   _fetch = fetchImplementation
 }
 
-export async function getZapEndpoint(metadata: Event): Promise<null | string> {
+export async function getZapEndpoint(metadata: Event<0>): Promise<null | string> {
```
alexgleason, line 16: This is the only potentially breaking change. TypeScript users will have to update their code to: `getZapEndpoint(event as Event<0>)`.

Once we update relay.ts to infer types from filters (in a later MR) this shouldn't be necessary.

.eslintrc.json
fiatjaf: Yes, please. eslint is stupid.

PR #207, opened 2023-05-07
nip05 refactoring

alexgleason:
Greatly simplify the nip05 lookup function with a good regex. Also parse the json response to make sure it's valid


fiatjaf:
This is good, but you changed the return type of the response. The `relays` field is just an array of strings, not an object.

alexgleason:
You're looking at the parsed nostr.json. The function still returns the same format (added a test)

fiatjaf:
Sorry, I'm stupid.

PR #214, opened 2023-05-12
Replace array list queue with a linked list one 

Egge7:
The relay object uses incomingMessageQueue as a way to keep a fifo queue of messages incoming from open ws connections. This array is a fundamental part of the whole library and its .unshift() gets called a lot.

Because JavaScript Arrays are ArrayLists, their .unshift() method is work intensive (O(n)). When called every single remaining item in it needs to get a new index assigned, which makes them a suboptimal choice for queues.

I believe we should replace the standard ArrayList with data structure more suited for this task. I have added a MessageQueue class to the ulitities that implements a LinkedList, which exposes enqueue and dequeue methods(both O(1)).

Before I spent more time on this I wanted to get the maintainers input on this. Is there maybe something that I am missing that made the standard array the better choice?

### DIFF
--- a/relay.ts
+++ b/relay.ts
@@ -3,6 +3,7 @@
 import {verifySignature, validateEvent, type Event} from './event.ts'
 import {matchFilters, type Filter} from './filter.ts'
 import {getHex64, getSubscriptionId} from './fakejson.ts'
+import { MessageQueue } from './utils.ts'
 
 type RelayEvent = {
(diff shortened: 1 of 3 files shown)


fiatjaf:
I don't know if this is really good performance-wise, but I'm convinced, let's do it!

Egge7:
This should be good to go

PR #220, opened 2023-05-26
NIP-26: Allow multiple kinds in createDelegation

enjikaka:
The NIP-26 doc says multiple kinds are valid in the cond. This modifies the createDelegation method to allow putting an array of numbers for the kind param.

### DIFF
--- a/nip26.test.ts
+++ b/nip26.test.ts
@@ -103,3 +103,13 @@ test('create and verify delegation', async () => {
   })
   expect(getDelegator(event)).toEqual(pk1)
 })
+
+test('can create delegation with multiple kinds', async () => {
+  let sk1 = generatePrivateKey()
+  let sk2 = generatePrivateKey()
+  let pk2 = getPublicKey(sk2)
+  let delegation = createDelegation(sk1, {pubkey: pk2, kind: [1, 4]})
+
+  expect(delegation.cond).toContain('kind=1')
+  expect(delegation.cond).toContain('kind=4')
+})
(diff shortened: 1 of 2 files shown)


fiatjaf:
I need someone who understands NIP-26 to vouch for this.

pablof7z:
this is correct, but the chances that this will work are slim

we should put NIP-26 out of it's misery

enjikaka:
> we should put NIP-26 out of it's misery

We were planing on using delegation in a nostr solution to post on users behalf instead of having them pass out their private key. But if it is going away we should abort that work. 

What's the plan for NIP-26 and the reasoning to pulling the plug on it, @pablof7z? :)

arthurfranca:
Taking advantage of the ongoing discussion.. what do you think of this login workflow?

1) Site X creates to user a privkey (so site already knows the privkey)
2) User copies privkey
3) Now that user has it, instead of the site keeping the privkey locally, which is a security risk, it creates and stores a NIP-26 delegation (with multiple kinds rights it needs to work, like this PR enables) that has a duration of say 15 days.
4) Site has now a session with 15 days duration. After that time passes, user has to paste the privkey again (no problem because the site was the one who created it in the first step)

Is this a legit client-side login workflow?

arthurfranca:
Or do you think its better to have a ZBD-like approach with a custodial privkey (server-side)?

But then if the user wants to reuse the key at another client, instead of showing it the privkey, ZBD-like client shows a 1-day delegation token the user can use at the other client? So we create this agreement between clients of using 1-day delegation tokens. 🤔️

fiatjaf:
NIP-26 has been removed from this library.

PR #221, opened 2023-05-31
Implement NIP-44: secure versioned replacement for NIP4

paulmillr:
https://github.com/nostr-protocol/nips/pull/574

### Benchmark

```
nip4 x 19,419 ops/sec @ 51μs/op
nip44 x 101,235 ops/sec @ 9μs/op
nip44 csv x 108,660 ops/sec @ 9μs/op
```


earonesty:
on the massive efficiency gains alone, we should switch to this.   we already switched to it on our group chat, and it blazes.

earonesty:
> @earonesty do you feel like it's better than aes solution? Because i'm not 100% sold on xchacha with one key instead of chacha/aes with different keys

rotating keys feels safer, but also xchacha20's simpler construction makes it easier to see that it should be safe with 1 key

paulmillr:
xchacha20 is basically using `hchacha` internal hash to convert `192-bit nonce` and `key` to chacha's `96-bit nonce` & different `key`. So it's actually using separate chacha key for every message, internally.

As i've been thinking more about this, the main issue I see is the fact iOS doesn't ship xchacha (it ships chacha20-poly1305). Neither does Android. Neither will webcrypto (in 2025? or whenever it'll be shipped). Also, xchacha doesn't have a spec, it had a draft spec, but it's expired now. xchacha cannot be built using chacha20-poly1305 and adding hchacha on top of it: from my tests, presence of poly1305 affects the behavior.

So while xchacha is still reasonable, maybe we re-evaluate our options.

And you're absolutely correct in reasoning about secp256k1 performance: it's *the* bottleneck, 30-100x slower than chacha or aes.

earonesty:
at this point im more concerned with the metadata on DM's.   using a tweaked-private-key results in a massively better ux than ephemeral public gift-wrap style keys for DM's because the client can remain stateless.   my only question is "plausibly-deniable but forgeable between parties" vs "unforgeable but the recipient can prove you sent it"

i'm seriously considering making a "plausibly deniable" slider in the app and supporting both.   which feels kinda crazy, but there's no way around that tradeoff in real life.

alexgleason:
Plausible deniability already doesn't exist on Nostr. So much time has been wasted on ActivityPub on the same issue. A benefit of Nostr is that a decision was already made to do this the easy way instead of the hard way.

earonesty:
i call the impl "44X": https://github.com/ArcadeLabsInc/arclib/pull/35/files

it's not terribly hard,  just signs with the shared secret as opposed to signing with your private key.

paulmillr:
@fiatjaf ready 2 merge

staab:
Do we need the updates to nip44? The code is not as easy to follow, and `ensureBytes` is redundant with the typescript annotations.

paulmillr:
I don't think we need to add ensureBytes. A simple sanity check (that everything else is the same) would suffice.

paulmillr:
Updated

paulmillr:
Let's merge this one in @fiatjaf

fiatjaf:
Sorry, I need someone to update me on what is happening. I am very confused about what is happening from reading the NIP-44 discussions.

paulmillr:
@fiatjaf here is the relevant, updated specification that describes everything: https://github.com/nostr-protocol/nips/blob/ba8c0101300f7ef4c134b81941db0fdb79878040/44.md

fiatjaf:
I will still read it calmly, but looks like it's good. If you prefer I can just merge this right now, or is it ok if I take some more days? Not that my inputs will make any difference.

paulmillr:
@fiatjaf let's merge it

PR #239, opened 2023-06-29
Fix nip27.matchAll crash on invalid nip19

alexgleason:
I received this event containing an invalid nip19: https://www.nostr.guru/e/bd107ef2033ff82a42e5a1377020771253d795568a6db0e2b53ad53cb35d73d7

It crashed my web server, which uses the `nip27.matchAll` function.

I wrote that function. I thought a try-catch wasn't needed since we do a regex match first, but this nip19 matches the regex and still crashes because it's invalid (I think). So now we add a try-catch.

I'm not sure if it should crash or not, or what this nip19 is trying to represent. But this is what caused it:

```
npub129tvj896hqqkljerxkccpj9flshwnw999v9uwn9lfmwlj8vnzwgq9y5llnpub1rujdpkd8mwezrvpqd2rx2zphfaztqrtsfg6w3vdnljdghs2q8qrqtt9u686
```

### DIFF
--- a/nip27.test.ts
+++ b/nip27.test.ts
@@ -29,6 +29,25 @@ test('matchAll', () => {
   ])
 })
 
+test('matchAll with an invalid nip19', () => {
+  const result = matchAll(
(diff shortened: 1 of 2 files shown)


fiatjaf:
I think this is the right solution. nip19 throwing, but nip27 catching it and ignoring. JavaScript makes it hard to know what is going to throw and what isn't.

alexgleason:
>  JavaScript makes it hard to know what is going to throw and what isn't.

:frowning_face: https://github.com/microsoft/TypeScript/issues/13219#issuecomment-1515037604

PR #248, opened 2023-07-13
fix nip27 test

zappityzap:
This fixes the failing nip27 test. I have no idea why. Typescript hurts my brain.

PR #249, opened 2023-07-13
Fix nip27 type

alexgleason:
Supersedes #248

### DIFF
--- a/nip27.ts
+++ b/nip27.ts
@@ -13,7 +13,7 @@ export interface NostrURIMatch extends NostrURI {
 }
 
 /** Find and decode all NIP-21 URIs. */
-export function* matchAll(content: string): Iterable<NostrURIMatch> {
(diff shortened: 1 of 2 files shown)


nip27.ts
```diff
@@ -56,7 +56,7 @@ export function replaceAll(
   content: string,
   replacer: (match: NostrURI) => string
 ): string {
-  return content.replaceAll(regex(), (uri, value) => {
+  return content.replaceAll(regex(), (uri, value: string) => {
```
alexgleason, line 59: The type of value was `any` due to the type of `replaceAll`, causing the top function overload to match. Making it a string (the correct value based on the regex) makes it match the correct overload.

alexgleason:
Tests are still failing for an unrelated reason. Annoying, but this is an improvement so I'll merge.

PR #257, opened 2023-07-23
fix breaking change

utxo-one:
Fixes this issue which has been driving me crazy

happens in react, nextjs and svelte using the latest build packs.

React Error: (webpack)
(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.

Svelte Error (vite)

image


utxo-one:
i see someone else already has an approved PR that fixes this.

PR #258, opened 2023-07-23
Add WebSocket polyfill for Node.js

zeevo:
I am seeing crashes when running the README examples on `node` (v18.12.1).

```
/Users/zeevo/Projects/nostr/node_modules/nostr-tools/lib/nostr.cjs.js:473
        ws = new WebSocket(url);
        ^

ReferenceError: WebSocket is not defined
    at /Users/zeevo/Projects/nostr/node_modules/nostr-tools/lib/nostr.cjs.js:473:9
    at new Promise (<anonymous>)
    at connectRelay (/Users/zeevo/Projects/nostr/node_modules/nostr-tools/lib/nostr.cjs.js:471:25)
    at Object.connect (/Users/zeevo/Projects/nostr/node_modules/nostr-tools/lib/nostr.cjs.js:579:11)
    at /Users/zeevo/Projects/nostr/src/sub.js:22:17
    at Generator.next (<anonymous>)
    at /Users/zeevo/Projects/nostr/src/sub.js:8:71
    at new Promise (<anonymous>)
    at __awaiter (/Users/zeevo/Projects/nostr/src/sub.js:4:12)
    at /Users/zeevo/Projects/nostr/src/sub.js:21:8

Node.js v18.12.1
```

Polyfill'ing `WebSocket` in `relay.ts` seems to fix this.

### DIFF
--- a/relay.ts
+++ b/relay.ts
@@ -1,4 +1,5 @@
 /* global WebSocket */
+import 'websocket-polyfill'
 
 import {verifySignature, validateEvent, type Event} from './event.ts'
 import {matchFilters, type Filter} from './filter.ts'


alexgleason:
Man, nodejs really sucks. If this is a new project I highly recommend using Deno instead. This library is meant to be used in the browser as well, and I'm not sure about polyfilling it since it would bloat the size of the library in a browser.

zeevo:
Generally, I think we should use [isomorphic-ws](https://www.npmjs.com/package/isomorphic-ws) (5mm weekly downloads) to achieve this instead of `websocket-polyfill`.

If this sounds like a good idea to maintainers, then I volunteer to implement it.

alexgleason:
It will still load the entire polyfill in a browser tho unless you chunk it as part of the build process. I didn't realize we were already facing that with a websocket polyfill.

alexgleason:
Polyfills should generally be added by the application itself, not the library. That way the application can choose to load it async conditionally, etc. So unless we provide two builds, one with polyfills and one without, I would actually argue that the WS polyfill should just be removed.

zeevo:
@alexgleason I hear ya. [isomorphic-ws](https://www.npmjs.com/package/isomorphic-ws) claims it is only 4kB (compared to the 23.7 kB of [websocket-polyfill](https://www.npmjs.com/package/websocket-polyfill). Perhaps, a small price to pay for Node support?

zappityzap:
I would like to see isomorphic-ws as well: https://github.com/nbd-wtf/nostr-tools/issues/259

alexgleason:
So, `websocket-polyfill` is a devDependency actually, and is only used by tests. I was almost going to agree with you, but this changes my opinion. The readme says this:

>To use this on Node.js you first must install `websocket-polyfill` and import it:
>
```js
import 'websocket-polyfill'
```

isomorphic-ws does indeed look a lot more sane, but I still don't think it's worth it, because:

1. It's easy to work around in Node.js
2. It's actually a pro rather than a con that you have the flexibility to decide which polyfill to use on your end.
3. A core tenant of this library is that it only relies on two dependencies for cryptography:

> Only depends on `@scure` and `@noble` packages.

Adding the dependency here can actually interfere with existing projects that use their own polyfill. In my view the problem here is Node.js. The fact it doesn't implement web APIs is only one of its problems: https://deno.com/blog/commonjs-is-hurting-javascript

d-r-w:
Just from my perspective, the current docs stating that the polyfill is needed in node was fine.

Installing the polyfill as a dependency and importing it before using nostr-tools worked as expected.

From a design perspective, perhaps a balanced solution is having an overloaded way of requiring this dependency in the code if a node environment is detected. In this way, using implementing in node is more straightforward and removes the need for a proper import order.

All that being said, this was not an issue for me as the documentation is sufficient.

Issue #262, opened 2023-07-30
Need async/await help accessing relays


teknari:
Hello,

Using async/await to interact with a relay that I am connected to. For example, I can do the following when requesting an event:

```
let FindEvent = await relay.get({ ids: [96b826...] })
console.log(FindEvent)
```

The await "halts" until the relay.get is complete, and then prints out the event on the next line.

I would like to do something similar when I publish an event. Publish the event, wait for result (published or failed) and then print it out the result. Maybe something like this:

```
let pub = await relay.publish(Event)
console.log(pub)
```

This doesn't work however. The program flows past the await before the event is published.

Is there a way to accomplish this flow?

Thanks

d-r-w:
How does this work when publishing to a pool? I am using the API in the same way but publishing to a pool instead of to a single relay - awaiting SimplePool.publish[] gives me this error:

`Error: blocked: no active subscription
    at Timeout.handleNext [as _onTimeout] ([...]/nostr-tools/lib/nostr.cjs.js:552:27)
    at listOnTimeout (node:internal/timers:573:17)
    at process.processTimers (node:internal/timers:514:7)`
    
   I get this both when awaiting a Promise.all() and when just awaiting each sequentially.
    
 I see this comment as part of the commit, could it be clarified any?:
>we used to need the event emitter thing because we were subscribing to the same relay
to check if the event had been published, but that is not necessary now that we assume
an OK response will always come.

For my purposes, I'm more interested in knowing that the message was _sent_ successfully (in a try/catch) vs whether or not it's seen on the relay.

PR #263, opened 2023-07-30
Turn `.publish()` into a normal async function returning a promise

fiatjaf:
this simplifies the code and makes the API more intuitive.

we used to need the event emitter thing because we were subscribing to the same relay to check if the event had been published, but that is not necessary now that we assume an OK response will always come.

closes https://github.com/nbd-wtf/nostr-tools/issues/262

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,13 +108,7 @@ let event = {
 event.id = getEventHash(event)
 event.sig = getSignature(event, sk)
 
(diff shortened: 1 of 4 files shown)


alexgleason (approved):
This is awesome.

alexgleason:
You have a few failing tests is all.

PR #264, opened 2023-07-31
Create nip28.ts and nip28.test.ts

Airtune:


Airtune:
Implementation of https://github.com/nostr-protocol/nips/blob/master/28.md

Has automated tests in nip28.test.ts but is untested in production.

PR #267, opened 2023-08-08
+nip98.unpackEventFromToken +nip98.validateEvent

Airtune:


alexgleason (approved):
Beautiful diff. 

cc @Dolu89

Dolu89 (changes requested):
Can you please update the example? It's an error from me, it should be like this
```
await nip98.getToken('https://example.com/login', 'post', (e) =>sign(e), true);
```

Airtune:
> Can you please update the example? It's an error from me, it should be like this
> 
> ```
> await nip98.getToken('https://example.com/login', 'post', (e) => sign(e), true);
> ```
> 
> EDIT: on file nip98.ts line 23

I updated the example and also updated the example for validation on line 63:

```
await nip98.validateToken('Nostr base64token', 'https://example.com/login', 'post')
```

Dolu89 (approved):
Nice update

PR #268, opened 2023-08-08
Add NIP-15 Event kinds

natoe33:


natoe33:
@fiatjaf

natoe33:
@alexgleason

alexgleason:
The `Kinds` enum is deprecated. You should just use the number.

PR #272, opened 2023-08-12
Run auto formatter

staab:
Tests also aren't passing, are they set up to run against the actual network? Seems like that could be the source of errors.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -154,7 +154,7 @@
   "overrides": [
     {
       "files": ["**/*.test.ts"],
-      "env": { "jest/globals": true },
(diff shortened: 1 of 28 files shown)


alexgleason:
I like the idea of enforcing code style, but we need to make the CI fail when the code style doesn't match, or else this is an endless cat and mouse game.

I'm also not a fan of the line length limit that's forcing a bunch of code onto multiple lines. I think that should be increased.

staab:
Yeah, very true. I didn't change anything here, I just ran `yarn format`. I'm not really interested in fixing it for real, so I'll close unless someone wants to take over.

PR #278, opened 2023-08-15
nip44 updates

staab:
https://github.com/nostr-protocol/nips/pull/715

Take two, this time using a custom TLV. Note: I've never used TLVs before, so a review to make sure it's correct and idiomatic would be appreciated.

### DIFF
--- a/nip44.ts
+++ b/nip44.ts
@@ -9,12 +9,11 @@ import {utf8Decoder, utf8Encoder} from './utils.ts'
 export const getSharedSecret = (privkey: string, pubkey: string): Uint8Array =>
   sha256(secp256k1.getSharedSecret(privkey, '02' + pubkey).subarray(1, 33))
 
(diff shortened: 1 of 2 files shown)


staab:
It does now

staab:
Ok, updated this, tests should be passing now.

nip44.ts
```diff
@@ -25,32 +19,29 @@ export function encrypt(privkey: string, pubkey: string, text: string, v = 1) {
...
   }
 
-  if (data.v !== 1) {
-    throw new Error('NIP44: unknown encryption version')
+  if (data[0] !== 1) {
+    throw new Error(`NIP44: unknown encryption version: ${data[0]}`)
   }
 
+  const nonce = data.slice(1, 25)
+  const ciphertext = data.slice(25)
```
fiatjaf, line 42: `slice` is bad because it copies, `subarray` doesn't copy, so it's faster

PR #279, opened 2023-08-21
add batchedList method to SimplePool

sepehr-safari:
### Diff
1. added a new optional parameter `batchInterval` to SimplePool class `Default: 100ms`
2. added a new method `batchedList` to SimplePool class: it's similar to `list()` but with a batching mechanism.
3. fix a bug to filter duplicate relays in `sub()`

### `batchedList()` gets three parameters:
1. batchKey as a string: used to group different subscriptions together into a single websocket request.
2. relays: same as `list` method
3. filters: same as `list` method

### best practices:
1. combine same relays and similar filters together with a unique batchKey e.g. `authors` `notes` `contacts` `reactions` it's totally up to you
2. use batching for repetitive scenarios like getting the details of multiple profiles at the same time with the same and unique batchKey e.g. `authors`
3. create a single SImplePool instance in your app wide and use it in different components at the same time. you can use a global state manager (on frontend frameworks eg React/Vue) based on your needs.

### Example:
``` *.tsx

const App = () => {
  // subscribe to some relays and get some notes with kind 1
  const latestNotes = pool.list(relays, { kinds: [1], limit: 20 });
  // you don't have full details of the authors yet, just their pubkey
  // you will let each note component fetch its author details on its own

  return (
    <div>
      {latestNotes.map((note) => (
        <div key={note.id}>
          <NoteCard note={note} />
        </div>
      ))}
    </div>
  );
};

const NoteCard = ({ note }: { note: Note }) => {
  // fetch the author details with batchedList method with key 'authors'
  const author = pool.batchedList('authors', relays, {
    kinds: [0],
    limit: 1,
    authors: [note.author],
  });
  // SimplePool will batch all subscriptions with the same key together
  // and send JUST ONE request to each relay for each key
  // and then it will distribute the corresponding events to each subscription
  // so you can subscribe to the same key in different components
  // and you will get JUST DESIRED events for each component separately

  return (
    <div>
      <div>{note.content}</div>
      <div>{author.name}</div>
    </div>
  );
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "nostr-tools",
-  "version": "1.14.1",
+  "version": "1.14.2",
   "description": "Tools for making a Nostr client.",
   "repository": {
     "type": "git",
--- a/pool.ts
+++ b/pool.ts
@@ -7,26 +7,38 @@ import {
 import {normalizeURL} from './utils.ts'
 
 import type {Event} from './event.ts'
-import type {Filter} from './filter.ts'
+import {matchFilters, type Filter} from './filter.ts'
+
+type BatchedRequest = {
+  filters: Filter<any>[]
+  relays: string[]
+  resolve: (events: Event<any>[]) => void
+  events: Event<any>[]
+}
+
 export class SimplePool {
   private _conn: {[url: string]: Relay}
   private _seenOn: {[id: string]: Set<string>} = {} // a map of all events we've seen in each relay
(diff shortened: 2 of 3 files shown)


sepehr-safari:
tnx @fiatjaf would you also consider publishing this version?

fiatjaf:
Does it work?

sepehr-safari:
> Does it work?

Y. like a butter 🤌🏻

fiatjaf:
Published.

PR #281, opened 2023-08-24
Fix typescript definitions

### LINKED ISSUE #255: Build fails on import statement after v1.10.0 "nostr-tools has no exported member"
Upgrading to v1.11 or higher results in a build time error when importing from nostr-tools. 

Example: 

```ts
import { SimplePool } from "nostr-tools";
```

Results in:

```bash
 error TS2305: Module '"nostr-tools"' has no exported member 'SimplePool'
```

The same happens with other imports from nostr-tools too. 

Possibly related to: https://github.com/nbd-wtf/nostr-tools/commit/9f896479d0758886278af91f82afe649bce4401f

wbobeirne:
Closes #255. I'd love if @alexgleason weighed in since this reverts the changes from #213 that apparently were for Deno users. It'd be also nice to hear if this fixes the build woes @steliosrammos or @rolznz were experiencing.

### What This Does

* Removes the `allowImportingTsExtensions` configuration in `tsconfig.json`
  * Updates all ts files to remove `.ts` extension from imports
* Removed instances of invalid syntax `import { type Type } from ...`
* Some minor prettier diffs due to running `yarn format`

Some tests are failing, but it appears the test suite has been failing on master for quite some time.


rolznz:
My issue was resolved by upgrading Typescript to 5.X. But these seem like good improvements, and more standard usage of Typescript.

fiatjaf:
We should have stuck with just JavaScript.

wbobeirne:
> Is there a reason why people wouldn't want to upgrade their TypeScript?

Not a particularly good one. Maybe you're just relying on deprecated configs, or you're in an organization that uses a single version of TypeScript across a monorepo and you don't have the buy-in of everyone to upgrade right now.

Like I said if y'all just want to forge ahead feel free to close this out I don't really care, but I'd recommend adding a note somewhere to give people a hint that they need to be on TypeScript 5 for types to work since the errors you get are unintuitive if you're on TypeScript 4.

alexgleason:
I ended up adding a note to the readme in #287 about TypeScript 5+ being required. I also fixed the lint/format situation and made the CI fail if the format is invalid.

If you have a project where you cannot upgrade to 5.x and you're actively experiencing issues, I am open to the other changes in here.

wbobeirne:
Fine by me, appreciate ya taking a look.

PR #283, opened 2023-08-27
Fix code samples in README

jiftechnify:
* Introduce `finishEvent()`.
  + "Creating, signing and verifying events" sample is left untouched because the purpose is illustrating event creation step by step.
* Deal with the signature change of `SimplePool#publish()`.
* Add the `SimplePool#close()` usage.
* Fix the description about configuring `node-fetch`.
  +  In Node.js v18 or later you don't have to use `node-fetch` since "native" `fetch` is available by defeault.
* Make callbacks which use `await` keyword in it `async`.


alexgleason:
Wonderful

PR #284, opened 2023-08-29
Added eoseSubTimeout to pool's SubscriptionOptions

Egge7:
This allows SimplePool users to specify a custom eoseSubTimeout for a single subscription instead of for the whole pool instance.


Egge7:
Maybe there should be a way to disable eoseSubTimeout for SimplePool.sub all together (maybe by passing 0 as value)? So that EOSE can be trusted in this case.

PR #287, opened 2023-08-31
Fix tests, format everything with prettier and enforce prettier+eslint in the CI

alexgleason:
Introduces `just lint` to check if the files are all good, and `just format` now also formats with eslint.

I increased the line-length of prettier to a more roomy 120 characters. It now also wants spaces around objects/imports, which is the most opinionated change I made, but it's a lot more common in JS and what I think people are used to.

Fixes tests so the CI passes again, and now the CI will enforce code formatting with `just lint`.


alexgleason:
Man, these pool tests are terrible. Not sure how to fix them. I think we should still merge this.

PR #288, opened 2023-09-02
Draft: Add a symbol to verified events

alexgleason:
This is an experiment with types:

```ts
import { verifiedSymbol, verifySignature, type Event } from 'nostr-tools'

const event: Event = { ... };

console.log(event[verifiedSymbol]) // `undefined` - the event has not been verified or rejected

verifySignature(event) // true

console.log(event[verifiedSymbol]) // `true` - the event has now been verified

verifySignature(event) // true, returns immediately instead of calling `schnorr.verify`
```

Now calling `verifySignature(event)` will set a `[Symbol(verified)]: true` on the event, so calling `verifySignature` multiple times on the same event will have no effect.

You can also guard functions to only allow verified events:

```ts
function validateAuthEvent(event: VerifiedEvent): event is VerifiedEvent<27235> {
  // do stuff with the verified event...
}

const event: Event = { ... }

validateAuthEvent(event) // type error
validateAuthEvent(verifySignature(event)) // pass
```

A good thing about symbols is that they're unique to the session, so you cannot spoof them. You have to import the symbol from nostr-tools to use this functionality directly, and the symbol will be dropped from `JSON.stringify`.

### DIFF
--- a/event.ts
+++ b/event.ts
@@ -5,6 +5,9 @@ import { bytesToHex } from '@noble/hashes/utils'
 import { getPublicKey } from './keys.ts'
 import { utf8Encoder } from './utils.ts'
 
+/** Designates a verified event signature. */
+export const verifiedSymbol = Symbol('verified')
+
 /** @deprecated Use numbers instead. */
 /* eslint-disable no-unused-vars */
 export enum Kind {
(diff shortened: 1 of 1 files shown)


alexgleason:
The main problem this solves is that `verifySignature` is computationally expensive, but sometimes you want a function to guarantee that the event is verified from any source, especially when building libraries. Sometimes you need to build layers where the signature is valid at every step.

This allows you to call it as many times as you want in your code while expending the effort only once.

fiatjaf:
This is very cool. Does it make sense to do the same for hashing the event into its id?

alexgleason:
> This is very cool. Does it make sense to do the same for hashing the event into its id?

That's a good point. I think `[verifiedSymbol]: true` should mean the id and sig are both valid, rather than introducing a new symbol for the id.

Fortunately the `verifySignature` function already recomputes the event ID instead of just trusting `event.id`. But it doesn't force the id to actually be valid.

We should either make `verifySignature` mutate the id if it's wrong, or return `false`. Probably the latter.

PR #293, opened 2023-09-10
nip13: use a simpler implementation

alexgleason:
When I created the nip13 module, for some reason I glossed over the fact there was already a JavaScript implementation within the NIP-13 text itself: https://github.com/nostr-protocol/nips/blob/master/13.md#validating

I had pulled the implementation from Nostream, which is for Node.js, and adapted it to work in the browser. But apparently if you just do it the browser way it's much simpler.


alexgleason:
@Cameri can you take a look at this?

alexgleason:
This actually will work just fine in Node.js too.

PR #294, opened 2023-09-10
nip13: add minePow function

alexgleason:
Adds a function to mine the POW for an event. It's loosely adapted from Snort but I changed it quite a bit.

### DIFF
--- a/nip13.test.ts
+++ b/nip13.test.ts
@@ -1,7 +1,21 @@
-import { getPow } from './nip13.ts'
+import { getPow, minePow } from './nip13.ts'
 
 test('identifies proof-of-work difficulty', async () => {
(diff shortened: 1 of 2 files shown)


nip13.test.ts
```diff
@@ -1,7 +1,21 @@
...
+  const event = minePow({
+    kind: 1,
+    tags: [],
+    content: 'Hello, world!',
+    created_at: 0,
+    pubkey: '79c2cae114ea28a981e7559b4fe7854a473521a8d22a66bbab9fa248eb820ff6',
+  }, difficulty)
+
+  expect(getPow(event.id)).toBeGreaterThanOrEqual(difficulty)
+})
```
alexgleason, line 21: It will take less than 1 second to mine an event to this difficulty in the test runner. In my tests it took between 20ms and 40ms every time.

PR #295, opened 2023-09-11
Improve example for finishEvent

AsaiToshiya:
Remove assignment to `pubkey` from declaring an event because `pubkey` also assigned by `finishEvent`.


AsaiToshiya:
@alexgleason thanks for the quick merge!

PR #297, opened 2023-09-15
Update README.md

sepehr-safari:
add docs about batchedList.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,11 +144,17 @@ let event = await pool.get(relays, {
   ids: ['44e1827635450ebb3c5a7d12c1f8e7b2b514439ac10a67eef3d9fd9c5c68e245'],
 })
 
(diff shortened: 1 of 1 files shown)


fiatjaf:
This is good, because I tried but didn't understand how it works.

sepehr-safari:
it has a pretty simple mechanism.

when you execute the `batchedList` function with a specific `batchKey` e.g. `Authors`, it will wait a fixed amount of time (default: 100ms) before sending the ws request. if you execute another function call in that period of time and also if both of them has the same `batchKey` string, the SimplePool with merge those filters into a single filters array and then send a single ws request instead of two.

main use case:
when you want to show a list of notes to the user, and you have to fetch more details for each individual note (e.g. author details, reactions, etc) you can just render all note components together and let each note component execute its own `batchedList` function with its specific filters and you can be pretty sure that just a single ws request will be fired.

PR #298, opened 2023-09-17
Update nip05.ts

242:
Added support for hieroglyphs and emoji domains for nip05 verification (242@𓂀.cc should pass)

### DIFF
--- a/nip05.ts
+++ b/nip05.ts
@@ -7,7 +7,7 @@ import { ProfilePointer } from './nip19.ts'
  * - 1: name (optional)
  * - 2: domain
  */
-export const NIP05_REGEX = /^(?:([\w.+-]+)@)?([\w.-]+)$/
+export const NIP05_REGEX = /^(?:([\w.+\-]+|[\p{Lo}\p{Emoji}]+)@)?(([\w\-]+|[\p{Lo}\p{Emoji}]+)+(\.[\w\-.]+)*)$/u
 
 var _fetch: any
 


alexgleason:
I like your idea! But I don't think this is the right place for it. `242@𓂀.cc` should not pass, although  `242@xn--ts7d.cc` should pass. It should be up to the client to parse punycode in URLs and render the emoji during display. Allowing anything except strict URL characters here could open the door to potential security issues.

242:
I understand your point completely! However, requiring clients to adopt this might compromise the functionality of NIP05 in many instances where the library is being used, mainly because a large number of users are unaware of this feature's existence.

As it stands, the criteria for URL characters remain stringent, permitting only the use of emojis and hieroglyphs that do not resemble or mimic standard characters, thereby preventing possible security vulnerabilities. 

Please reconsider 🙏🍄

242:
@alexgleason, @fiatjaf 
Additionally, as per the guidelines outlined in NIP-05, it is specified that:
`NIP-05 presumes that the <local-part> will strictly adhere to a character set comprising of a-z, 0-9, and the symbols -_., with no regard to case sensitivity.`

This restriction should exclusively apply to the <local-part> and not extend to the domain section of the identifier. Moreover, the responsibility of translating glyphs to punycode rests with the browser client, not with every other client being developed within the ecosystem.

Therefore, an identifier like `242@𓂀.cc` is in full compliance with the standards set forth in the NIP-05 document. 

It's worth noting that the regex pattern  `/^[A-Za-z0-9-_]+$/ `was established in an earlier iteration of nostr-tools and has been operational across various clients, with [metadata](https://metadata.nostr.com/#) being a notable example.
However, the recent modifications made to the regex pattern through the [pull request  207](https://github.com/nbd-wtf/nostr-tools/pull/207) have unfortunately led to compatibility issues in certain clients. It's vital to consider the implications of such changes to ensure a seamless user experience across all platforms.

242:
This fails every IDN domain for NIP-05 Verification.
Such as every .世界 domain. or anything with non-english characters.

alexgleason:
It's not just about security, but compatibility.

On the DNS/HTTP protocol level, `𓂀.cc` is not compatible, only `xn--ts7d.cc` is. I've already told you how to make it work. What is the blocker?

Clients will need to handle IDNs regardless, so it should be up to clients to display it correctly.

See also: https://en.wikipedia.org/wiki/Emoji_domain#Issues

> At present, only Punycode encoding is advised in e-mail addresses,[[31]](https://en.wikipedia.org/wiki/Emoji_domain#cite_note-31) e.g. "mail@xn--n28h.tld", as many MTAs and MUAs don't support [SMTPUTF8](https://en.wikipedia.org/wiki/SMTPUTF8)[[32]](https://en.wikipedia.org/wiki/Emoji_domain#cite_note-32) which would allow SMTP commands with non-ASCII characters.

alexgleason:
I'm open to the idea I'm wrong about this, but at the very least some other Nostr developers should chime in here on problems it would solve to support IDNs directly. And developers should be aware of this, because it creates two possible ways to represent the same nip05. It just seems like an unnecessary extra layer of complexity. The thing I don't like about it are the problems I haven't even anticipated.

242:
> It's not just about security, but compatibility.
> 
> On the DNS/HTTP protocol level, `𓂀.cc` is not compatible, only `xn--ts7d.cc` is. I've already told you how to make it work. What is the blocker?
> 
> Clients will need to handle IDNs regardless, so it should be up to clients to display it correctly.
> 
> See also: https://en.wikipedia.org/wiki/Emoji_domain#Issues
> 
> > At present, only Punycode encoding is advised in e-mail addresses,[[31]](https://en.wikipedia.org/wiki/Emoji_domain#cite_note-31) e.g. "[mail@xn--n28h.tld](mailto:mail@xn--n28h.tld)", as many MTAs and MUAs don't support [SMTPUTF8](https://en.wikipedia.org/wiki/SMTPUTF8)[[32]](https://en.wikipedia.org/wiki/Emoji_domain#cite_note-32) which would allow SMTP commands with non-ASCII characters.

I followed your advice and everything seems to work fine, thank you! 🙏 
However, I've noticed that the web browser itself takes care of the necessary punycoding, making additional client-side efforts redundant. This became clear to me when I was testing things on metadata.nostr.com network request and comparing it with my own setup.

While looking into [pull request 207](https://github.com/nbd-wtf/nostr-tools/pull/207), I realized that we might not need to limit this in the regex, especially if we want to stay compatible with NIP-05. I totally get that the thing you don't like about it are the things you haven't anticipated, but it seems this goes beyond just the technical aspects of unicode, punycode, ASCII or the internet.

The browser, acting as a sort of super-client, is already handling the necessary conversions. This means we don't really need to enforce these changes at the client level, which could make things a lot easier and straightforward for everyone involved.

It doesn't essentially lead to two representations of the NIP-05 on the backend, as it ultimately gets converted to the respective punycode regardless.

Thus, it solely affects the frontend, allowing developers the discretion to choose whether to display a 'xn--' sequence or a glyph/emoji. Hopefully, a consensus will emerge favoring the more visually appealing emojis over 'xn--' strings.

Additionally, if you attempt to ping 𓂀.cc, it functions correctly, and there's no requirement to personally convert it to punycode, since the OS's punycode.c handles that task automatically.

Remaining receptive to additional feedback from developers. 🥳

PR #299, opened 2023-09-19
Update getSignature

arthurfranca:
The problem is the below code calls getEventHash twice (the second time is inside getSignature)
```
  event.id = getEventHash(event) 
  event.sig = getSignature(event, privateKey) 
```

### DIFF
--- a/event.ts
+++ b/event.ts
@@ -53,7 +53,7 @@ export type EventTemplate<K extends number = number> = Pick<Event<K>, 'kind' | '
 export type UnsignedEvent<K extends number = number> = Pick<
   Event<K>,
   'kind' | 'tags' | 'content' | 'created_at' | 'pubkey'
->
+> & Partial<Pick<Event<K>, 'id'>>
 
 /** An event whose signature has been verified. */
 export interface VerifiedEvent<K extends number = number> extends Event<K> {
@@ -139,5 +139,5 @@ export function signEvent(event: UnsignedEvent<number>, key: string): string {
 
 /** Calculate the signature for an event. */
 export function getSignature(event: UnsignedEvent<number>, key: string): string {
-  return bytesToHex(schnorr.sign(getEventHash(event), key))
+  return bytesToHex(schnorr.sign(event.id ?? getEventHash(event), key))
 }


fiatjaf:
Maybe the best thing would be to have `getSignature()` set the `event.id`? That way we would ensure there is no mismatch between `event.id` and `getEventHash(event)`? But then we would need the function to have a different name, right?

What do you think, @alexgleason?

alexgleason:
I think the proposed change is decent. Sometimes UnsignedEvent does want to have an optional ID, like in the case of gift wrap and a few other NIPs. There's currently no way to represent that with `nostr-tools`, so allowing an optional ID could be beneficial.

I am just trying to think if there are any possible security issues this could raise, eg code that relies on `getSignature` doing the right thing even if the event ID is invalid... no, right? This function is for signing your own events, so it would mean your own code is wrong if the ID is wrong.

The other question is what people expect the behavior of `getSignature` should be, outside of `finishEvent`. There is hardly a reason to call this function manually, opposed to calling `finishEvent`, and I'm questioning why we even export it.

I think calling `getSignature` on an event that already has an ID is undefined behavior, and this MR is just changing one undefined behavior for a different undefined behavior, with a slight efficiency gain. So I reluctantly support this change.

Does anyone actually use `getSignature` on its own in the wild? I only did before I realized `finishEvent` exists.

alexgleason:
Also, adding a `getEventHash` function seems like a good way to go about it. I wouldn't export that function. Just for use by `finishEvent`.

fiatjaf:
That's true.

Maybe we should just inline `getSignature` into `finishEvent` and remove it.

alexgleason:
You don't have to inline it, just stop exporting it. We can remove the deprecated `signEvent` alias of it, too. Then this MR makes more sense because all it does is improve the efficiency of `finishEvent` with no changes to the user.

arthurfranca:
> just stop exporting it

What about the `minePow` function? It returns an event without a signature.

In this case, `finishEvent` would waste time recalculating pubkey and would wrongly rewrite the id.

`getSignature` is needed in that example.

alexgleason:
Oh, good point. Well in that case, the desired behavior of `getSignature` is to ignore the ID. With this MR, passing an event with an ID to `minePow` would make it run forever?

arthurfranca:
I was just saying there may be moments when one just needs a signature, not everything that finishEvent adds to the event.

arthurfranca:
Maybe the best thing for the lib users would be to export some raw helpers that getEventHash and getSignature internally use:

`export function getHash (string) { return bytesToHex(sha256(utf8Encoder.encode(string))) }`
`export function sign (hash, key) { return bytesToHex(schnorr.sign(hash, key)) }`

Because some use cases may require hashes of strings serialized differently than an event.id. The merged NIP-53 talks about a [proof](https://github.com/nostr-protocol/nips/blob/master/53.md#proof-of-agreement-to-participate) that basically is this:
```
const hash = getHash("<kind:pubkey:dTag>")
const signature = sign(hash, privkey) 
```

fiatjaf:
Then people are better just using `@noble/curves` directly.

If this library provides any value is by abstracting those things and providing specific functions. It needs a `nip53` module that generates those proofs from higher-level arguments.

PR #301, opened 2023-09-24
Draft: Use dynamic imports to chunk noble deps

alexgleason:
This is an experiment. Do not merge. It breaks everything by changing the functions into async functions.

Just an idea I had while thinking about #296. For some reason I thought the functions were already async, in which case this would be perfect. But since they aren't we can't do this.

By using dynamic imports, the bundler will create chunks, and it will actually wait for the browser to load the chunk before proceeding. This will improve the initial load time at the cost of the first call to the function being slower (certain tools can precache chunks, though).

It might actually be possible to do this for the async functions like `nip04.encrypt` and `nip04.decrypt`.

### DIFF
--- a/event.ts
+++ b/event.ts
@@ -1,7 +1,3 @@
-import { schnorr } from '@noble/curves/secp256k1'
-import { sha256 } from '@noble/hashes/sha256'
-import { bytesToHex } from '@noble/hashes/utils'
-
 import { getPublicKey } from './keys.ts'
 import { utf8Encoder } from './utils.ts'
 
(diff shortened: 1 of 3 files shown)


paulmillr:
this won't work in all environments, because `@noble/hashes/...` is unknown in a browser, etc.

alexgleason:
Right. It only works if the bundler supports it.

paulmillr:
Could you remind what's the problem here? If you don't use `generatePrivateKey`, its deps won't be bundled. If you use it, you will need it at some point, and by that point your bundler would be able to split deps into chunks without this pull request.

alexgleason:
> by that point your bundler would be able to split deps into chunks without this pull request.

Is this type of magic really possible? If so my bundles should be broken up a lot more. It only breaks where I use dynamic imports. https://fe.soapbox.pub/report.html

paulmillr:
Yeah, everyone does this (code splitting).

The nostr-tools themselves used `import *` which is not helpful to shaking, but maybe after the discussion from previous pull request, maybe that has been removed? If it wasn't, it's time to remove it and use imports like `import { generateprivatekey } from 'nostr-tools/event'`.

PR #302, opened 2023-09-24
edit batchedList and add mergeFilters to it

sepehr-safari:
just added mergeFilters to batchedList method


pool.ts
```diff
@@ -79,13 +79,10 @@ export class SimplePool {
     let eosesMissing = relays.length
 
     let eoseSent = false
-    let eoseTimeout = setTimeout(
```
sepehr-safari, line 82: there are also some auto code formats.

pool.ts
```diff
@@ -213,7 +210,7 @@ export class SimplePool {
               relays.push(...br.relays)
             })
 
-            const sub = this.sub(relays, filters)
+            const sub = this.sub(relays, [mergeFilters(...filters)])
```
sepehr-safari, line 213: this is the actual change. just added mergeFilters.

PR #305, opened 2023-09-26
Organize build, allow one entrypoint per file

fr4nzap:
Adds one entrypoint per file such that `import { generatePrivateKey } from 'nostr-tools/keys';` can be done from clients without having to import everything but the kitchen sink. For ESM and CJS, the IIFE version still bundles from `index.ts`

### DIFF
--- a/build.js
+++ b/build.js
@@ -2,9 +2,17 @@
 
 const fs = require('fs')
 const esbuild = require('esbuild')
+const { join } = require('path');
+
+const entryPoints = fs.readdirSync(process.cwd())
+  .filter(
+    (file) =>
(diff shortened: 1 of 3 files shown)


paulmillr:
lgtm

fiatjaf:
I don't know how to fix this conflict because I don't understand the meaning of these `package.json` fields.

fr4nzap:
@fiatjaf I changed it to `lib/types` (instead of `lib`) just to have it a bit more organized

Not sure about `resolveJsonModule`

> Allows importing modules with a .json extension, which is a common practice in node projects. This includes generating a type for the import based on the static JSON shape.

Maybe keep it and check that the whole thing still works

PR #313, opened 2023-10-24
Refactor mergeFilters() to eliminate the ts-ignores

WilliamConnatser:
I believe that this logic is effectively the same as what was originally implemented (but without the ts-ignores). Admittedly, my solution is slightly less performant than the original implementation, but in my opinion it's a minor trade off to ensure the function is strongly typed. If consumers of this library are being dumb by sending in huge arrays with a ton of redundant filters, then that's more user error than anything. In most realistic and efficiency-conscious scenarios, I think the performance trade-offs will be negligible.

### DIFF
--- a/filter.ts
+++ b/filter.ts
@@ -46,27 +46,41 @@ export function matchFilters(filters: Filter<number>[], event: Event<number>): b
 }
 
 export function mergeFilters(...filters: Filter<number>[]): Filter<number> {
-  let result: Filter<number> = {}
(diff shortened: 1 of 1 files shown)


fiatjaf:
Thank you for the PR and sorry for not answering this in time. Now there are too many conflicts to be worth merging, it's probably better to start anew.

But I think I personally prefer the ridiculously small performance gain over the type safety, so maybe let's not do this anyway.

PR #314, opened 2023-10-30
turn Kind enum into a simple object with numbers

fiatjaf:
Will this break something? Is it a good idea?

@alexgleason

### DIFF
--- a/event.ts
+++ b/event.ts
@@ -35,7 +35,7 @@ export const Kind = {
   BadgeDefinition: 30009,
   Article: 30023,
   FileMetadata: 1063,
-}
+} as const
 
 export interface Event<K extends number = number> {
   kind: K
--- a/event.ts
+++ b/event.ts
@@ -8,35 +8,33 @@ import { utf8Encoder } from './utils.ts'
 /** Designates a verified event signature. */
 export const verifiedSymbol = Symbol('verified')
 
-/** @deprecated Use numbers instead. */
-/* eslint-disable no-unused-vars */
-export enum Kind {
-  Metadata = 0,
-  Text = 1,
-  RecommendRelay = 2,
-  Contacts = 3,
-  EncryptedDirectMessage = 4,
-  EventDeletion = 5,
-  Repost = 6,
-  Reaction = 7,
-  BadgeAward = 8,
-  ChannelCreation = 40,
-  ChannelMetadata = 41,
-  ChannelMessage = 42,
-  ChannelHideMessage = 43,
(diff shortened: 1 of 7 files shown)


nip18.ts
```diff
@@ -23,7 +23,7 @@ export function finishRepostEvent(
   reposted: Event<number>,
   relayUrl: string,
   privateKey: string,
-): Event<Kind.Repost> {
+): Event {
```
alexgleason, line 26:
```suggestion
): Event<typeof Kind.Repost> {
```

alexgleason:
Yes, we should do this. It will break types for users, so it needs a major version bump probably. We should also preserve types on our own code. Just need to use `typeof`

alexgleason:
Also, this new object should be moved to `kinds.ts` instead of `events.ts`

event.ts
```diff
@@ -8,35 +8,33 @@ import { utf8Encoder } from './utils.ts'
...
+  Zap: 9735,
+  RelayList: 10002,
+  ClientAuth: 22242,
+  NwcRequest: 23194,
+  HttpAuth: 27235,
+  ProfileBadge: 30008,
+  BadgeDefinition: 30009,
+  Article: 30023,
+  FileMetadata: 1063,
 }
```
WilliamConnatser, line 38:
```suggestion
} as const
```

TBH I'd prefer that we stick with using an enum for this because I feel like it's the better tool for the job, but if we insist on changing this to a JavaScript object I'd suggest that we do a (const assertion)[https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions] so it is more strongly typed. This is the inferred type after the suggested change:
```
const Kind: {
    readonly Metadata: 0;
    readonly Text: 1;
    readonly RecommendRelay: 2;
    readonly Contacts: 3;
    readonly EncryptedDirectMessage: 4;
    readonly EventDeletion: 5;
    readonly Repost: 6;
    readonly Reaction: 7;
    ... 17 more ...;
    readonly FileMetadata: 1063;
}
```

The inferred type from your change is as follows:
```
const Kind: {
    Metadata: number;
    Text: number;
    RecommendRelay: number;
    Contacts: number;
    EncryptedDirectMessage: number;
    EventDeletion: number;
    Repost: number;
    Reaction: number;
    BadgeAward: number;
    ChannelCreation: number;
    ... 15 more ...;
    FileMetadata: number;
}
```

alexgleason: Good catch. Should be able to just add `} as const;` to the end of the whole object.

PR #320, opened 2023-11-28
Update `@noble/curves` to 1.2.0

yijiasu:
There was a bug in the package `@noble/curves` with version 1.1.0 when parsing strings with non-hex characters. It will not throw any error but keep processing with the wrong formatted string.

Here's the steps to re-produce:

```
const { getPublicKey } = require('nostr-tools')
const sk = "02336e32749f642c6a56ee709a01afe7ff8d2b268c04a7f;a0000308cd2382eb"
const pk = getPublicKey(sk)
```

The variable `sk` contains a non-hexadecimal character, I believe the function call `getPublicKey()` should throw an error here but it's keeping converting it to a public key.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -148,7 +148,7 @@
   "license": "Unlicense",
   "dependencies": {
     "@noble/ciphers": "0.2.0",
-    "@noble/curves": "1.1.0",
+    "@noble/curves": "1.2.0",
     "@noble/hashes": "1.3.1",
     "@scure/base": "1.1.1",
     "@scure/bip32": "1.3.1",


alexgleason (approved):
I wonder what kind of interesting security implications this might have. :thinking:

yijiasu:
The underlying reason of this bug was here:

https://github.com/paulmillr/noble-curves/blob/dba2f0e7321a8377d271d28ae7a7f5e482003080/src/abstract/utils.ts#L55C1-L62C1

Number.parseInt() will not return a `NaN` if the input string is beginning with a numerical char. It looks like it does not char if the string contains any other chars..

```
> Number.parseInt("f;", 16);
15
> Number.parseInt(";f", 16);
NaN
```

PR #324, opened 2023-12-14
Updating nip06 privateKeyFromSeedWords function 

laantungir:
## This pull request updates the nip06 function privateKeyFromSeedWords().

Previously, you were only allowed to get one private key from a seed phrase. I added the optional parameters "account", "change" and "index" which allows you to derive many private keys from one seed phrase.

The parameters are optional, and this change should be backward compatible.


ralyodio (approved):

fiatjaf:
NIP-06 only specifies `account`, the other parameters are hardcoded to `0`.

PR #328, opened 2023-12-20
Added bytesToHex for nsec and npub decoding

calvadev:
The data value for nsec was returning as an Uint8Array rather than a hex string because the bytesToHex conversion wasn't being applied.


fiatjaf:
But this is intentional! See the release notes: https://github.com/nbd-wtf/nostr-tools/releases/tag/v2.0.0

If you didn't see that you may wanna read it because other things have changed.

PR #340, opened 2024-01-03
Add getFilterLimit function

alexgleason:
When filtering events, it can sometimes be difficult to know when to stop. Obviously if the filter has a `limit` you can use it, but otherwise you may need to rely on timeouts or other cues.

But some filters have an intrinsic limit. For example, `{ ids: ["a", "b", "c" }` has an intrinsic limit is `3`.

If an intrinsic limit can be determined from the filter, `getFilterLimit` will return it. Otherwise it will return `Infinity`.

This is useful in multiple scenarios:

1. You can skip invalid filters entirely with `filters.filter((filter) => getFilterLimit(filter) > 0)`. This technique is usable on both clients and relays.
2. If you are querying events from a storage, you can cancel when the limit is reached, saving time.
3. You can use it to calculate the desired limit on your own filters before sending them to storages, which might help them EOSE properly in case they don't already implement a feature like this internally.

### DIFF
--- a/filter.test.ts
+++ b/filter.test.ts
@@ -1,5 +1,5 @@
 import { describe, test, expect } from 'bun:test'
-import { matchFilter, matchFilters, mergeFilters } from './filter.ts'
+import { getFilterLimit, matchFilter, matchFilters, mergeFilters } from './filter.ts'
 import { buildEvent } from './test-helpers.ts'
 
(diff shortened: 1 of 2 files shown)


fiatjaf:
This is an amazing breakthrough in the Theory of Nostr in the subfield of Filter Theory.

PR #341, opened 2024-01-05
handle connectionstrings with authority

Egge21M:
This is a tiny change in the NIP-47 utilities to handle the very real case of a NWC provider using connection URIs that don't comply with the spec and instead use "schema://pubkey". This change ensure compatibility with both.

Examples of NWC providers that use Authority are:

- Alby
- Current
- Mutiny

### DIFF
--- a/nip47.test.ts
+++ b/nip47.test.ts
@@ -10,9 +10,19 @@ import { NWCWalletRequest } from './kinds.ts'
 globalThis.crypto = crypto
 
 describe('parseConnectionString', () => {
-  test('returns pubkey, relay, and secret if connection string is valid', () => {
+  test('returns pubkey, relay, and secret if connection string is valid and complies with spec', () => {
     const connectionString =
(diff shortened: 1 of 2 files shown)


alexgleason:
If everybody is using the `//` in the URI and the spec doesn't match, maybe we should update the spec? And do it one way. Is there an MR to update the spec?

fiatjaf:
Am I stupid? I had never seen this PR here ever.

Egge21M:
> If everybody is using the `//` in the URI and the spec doesn't match, maybe we should update the spec? And do it one way. Is there an MR to update the spec?

While that does sound practical, I am not a fan of doing things wrong knowingly, just because the majority does it that way.

These implementations are in the wrong and should update. However I am also not going to die on this hill.

fiatjaf:
Is this PR still valid? I don't know anything about NWC, so please decide.

PR #345, opened 2024-01-09
Add NIP-40 module for event expiration

alexgleason:
Adds helper functions for dealing with event expiration.


alexgleason:
If an event is expired, matchFilters should arguably not return it. But I'm not brave enough to actually make that change, just some food for thought.

fiatjaf:
I'm not that brave either.

PR #350, opened 2024-01-17
Nip98 enhancement

sepehr-safari:
- added docs for all methods
- refactored with smaller and more readable functions
- added more test cases
- fixed possible bugs in checking tags:
`urlTag?.length !== 1 && urlTag?.[1] !== url`
that "length !== 1" doesn't make sense.
- fixed possible bugs in checking body:
`Boolean(body) && Object.keys(body).length > 0`
that would fail in non-object inputs.

### DIFF
--- a/nip98.test.ts
+++ b/nip98.test.ts
@@ -1,76 +1,83 @@
-import { describe, test, expect } from 'bun:test'
-import { getToken, unpackEventFromToken, validateEvent, validateToken } from './nip98.ts'
-import { Event, finalizeEvent } from './pure.ts'
-import { generateSecretKey, getPublicKey } from './pure.ts'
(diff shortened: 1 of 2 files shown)


alexgleason:
You may also find this useful: https://gitlab.com/soapbox-pub/ditto/-/blob/main/src/utils/nip98.ts?ref_type=heads 

Some of the code in this repo is hard to trust without zod, and I end up rewriting it.

sepehr-safari:
> You may also find this useful: https://gitlab.com/soapbox-pub/ditto/-/blob/main/src/utils/nip98.ts?ref_type=heads
> 
> Some of the code in this repo is hard to trust without zod, and I end up rewriting it.

I could help adding Zod to nostr-tools, generally, if it was planned.

PR #356, opened 2024-01-18
fix failing test cases

sepehr-safari:
- replaced relay addresses with `nos.lol` and fixed failing test cases

### DIFF
--- a/relay.test.ts
+++ b/relay.test.ts
@@ -3,7 +3,7 @@ import { afterEach, expect, test } from 'bun:test'
 import { NostrEvent, finalizeEvent, generateSecretKey, getPublicKey } from './pure.ts'
 import { Relay } from './relay.ts'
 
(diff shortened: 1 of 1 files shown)


fiatjaf:
The tests are flaky because relying on external relays for them is wrong, but they often work. Changing here won't help as they will eventually fail with nos.lol too.

A better approach could be to run the tests inside a docker thing that also spins up a nostr-rs-relay or something and then uses that. I don't know, I don't think it's that important.

sepehr-safari:
they won't fail with `nos.lol`.
by the way, I'm trying to implement mock websocket. will send a PR once it's ready.

fiatjaf:
Why won't they fail with nos.lol?

sepehr-safari:
I tested and it passed. it's about the response time of the relay. `nos.lol` is fast enough to pass the tests in our desired time.

PR #364, opened 2024-02-04
Fix explicit file extension for core.ts

quentintaranpino:
Some imports do not have the explicit extension and others do. In this case I fix core.ts which is giving error in some of my projects that use nostr-tools.

image


fiatjaf:
Thank you.

PR #365, opened 2024-02-06
Remove tsd package

asoltys:
This package doesn't appear to be used and seems to come with typescript 4 instead of 5 which gives me a tsc bin that can't build the project


fiatjaf:
Weird, I don't remember this package.

Issue #372, opened 2024-02-25
NIP49 is part of package but not exported by index.js / bundle


Egge21M:
nip49 is present in current versions, but can not be imported as its not exported.

fiatjaf:
No need to create issues before pull requests, please create just the issue or just the pull request.

PR #373, opened 2024-02-25
added export for nip49

Egge21M:
closes https://github.com/nbd-wtf/nostr-tools/issues/372

NIP49 is included in bundle, but can not be imported as its not exported.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -20,6 +20,7 @@ export * as nip39 from './nip39.ts'
 export * as nip42 from './nip42.ts'
 export * as nip44 from './nip44.ts'
 export * as nip47 from './nip47.ts'
+export * as nip49 from './nip49.ts'
 export * as nip57 from './nip57.ts'
 export * as nip98 from './nip98.ts'
 


fiatjaf:
I want to move to having all the imports be done as `import * as nip49 from 'nostr-tools/nip49'` instead of bundling stuff. Not adding new libraries to the bundle is part of nudging people into doing that.

Egge21M:
> I want to move to having all the imports be done as `import * as nip49 from 'nostr-tools/nip49'` instead of bundling everything always. Not adding new libraries to the bundle is part of nudging people into doing that.
> 

Oh got it! Thanks for pointing that out. Wouldn't modern bundlers tree shake anything that's not required anyways?

This is perfect however. Will close PR and issue.

fiatjaf:
> Wouldn't modern bundlers tree shake anything that's not required anyways?

I can't explain why, but in my experience no.

My best guess is that they're unable to tree-shake when you're importing from a file that is itself importing (with `*`) a bunch of other things (like [index.ts](../blob/master/index.ts)) because they don't know what kind of side-effects could be happening in these imports.

So we must import directly from these subfiles and never do the `*` in them.

PR #387, opened 2024-03-16
fix: error thrown on ws close

### LINKED ISSUE #370: Error: sending on closed connection
For some reason after a certain amount of time nostr-tools returns an error:
Error: sending on closed connection

I am leaving the connections open in order to fetch chats in real time. Is this not the way nostr-tools is intended to be used?

Thanks for your time btw and this package is great!

j3ko:
- Fix for an occasional error thrown when the websocket is closed.

fixes #370

### DIFF
--- a/abstract-relay.ts
+++ b/abstract-relay.ts
@@ -78,9 +78,9 @@ export class AbstractRelay {
     this.connectionPromise = new Promise((resolve, reject) => {
       this.connectionTimeoutHandle = setTimeout(() => {
         reject('connection timed out')
-        this.connectionPromise = undefined
         this.onclose?.()
         this.closeAllSubscriptions('relay connection timed out')
+        this.connectionPromise = undefined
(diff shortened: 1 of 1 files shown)


fiatjaf:
Thank you for finding the bug. I've reverted it and pushed a different fix that does the same thing, but in a way that I think is clearer. Please tell me if my fix is still buggy.

j3ko:
> Thank you for finding the bug. I've reverted it and pushed a different fix that does the same thing, but in a way that I think is clearer. Please tell me if my fix is still buggy.

Thanks, I haven't encountered any more errors after applying the new fix.  A similar change might be needed for the `connectionTimeoutHandle` block:

https://github.com/nbd-wtf/nostr-tools/blob/e30e08d8e2ed44c11092bf8b8a014e41b802624e/abstract-relay.ts#L81

PR #398, opened 2024-04-12
Add stable sortEvents function

alexgleason:
Adds a function to sort events, which is often needed when gathering events from untrusted sources such as relays.

Events are sorted reverse-chronologically by the `created_at` timestamp. If there is a tie with `created_at`, it sorts lexicographically by the event ID.

### DIFF
--- a/core.ts
+++ b/core.ts
@@ -49,3 +49,17 @@ export function validateEvent<T>(event: T): event is T & UnsignedEvent {
 
   return true
 }
+
+/**
+ * Sort events in reverse-chronological order by the `created_at` timestamp,
+ * and then by the event `id` (lexicographically) in case of ties.
+ * This mutates the array.
+ */
+export function sortEvents(events: Event[]): Event[] {
+  return events.sort((a: NostrEvent, b: NostrEvent): number => {
+    if (a.created_at !== b.created_at) {
+      return b.created_at - a.created_at
+    }
+    return a.id.localeCompare(b.id)
+  })
+}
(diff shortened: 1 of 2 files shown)


core.ts
```diff
@@ -49,3 +49,17 @@ export function validateEvent<T>(event: T): event is T & UnsignedEvent {
...
+ * Sort events in reverse-chronological order by the `created_at` timestamp,
+ * and then by the event `id` (lexicographically) in case of ties.
+ * This mutates the array.
+ */
+export function sortEvents(events: Event[]): Event[] {
+  return events.sort((a: NostrEvent, b: NostrEvent): number => {
+    if (a.created_at !== b.created_at) {
+      return b.created_at - a.created_at
+    }
+    return a.id.localeCompare(b.id)
```
alexgleason, line 63: @vitorpamplona re: https://github.com/nostr-protocol/nips/issues/759#issuecomment-1704539371 and https://github.com/nostr-protocol/nips/issues/759#issuecomment-1705335513, does my sort function match yours? I'm sorting by `created_at` desc, then by `id` asc. 

Are there any arguments for which direction to sort `id`? I guess events with proof of work would be sorted above others.

fiatjaf:
NIP-01 says this about replaceable events:

> In case of replaceable events with the same timestamp, the event with the lowest id (first in lexical order) should be retained, and the other discarded.

Does this match your function?

core.ts
vitorpamplona: I do both desc but I can change.

alexgleason:
Yes it does :+1: Cool.

PR #403, opened 2024-05-21
including interface for nip07

antonioconselheiro:
This pull request includes:
- typescript interface for nip07;
- centralized nip46 and nip07 relay map into a named type and export it;
- include docs of how to include the nip07 type for browser environment;

I haven't finished testing it yet and it's not ready, but I opened the pull request to advance the code review

### DIFF
--- a/nip07.ts
+++ b/nip07.ts
@@ -1,9 +1,9 @@
-import { EventTemplate, VerifiedEvent } from './core.ts'
+import { EventTemplate, NostrEvent } from './core.ts'
 import { RelayRecord } from './index.ts'
 
 export interface Nip07 {
   getPublicKey(): Promise<string>
-  signEvent(event: EventTemplate): Promise<VerifiedEvent>
+  signEvent(event: EventTemplate): Promise<NostrEvent>
   getRelays(): Promise<RelayRecord>
   nip04?: {
     encrypt(pubkey: string, plaintext: string): Promise<string>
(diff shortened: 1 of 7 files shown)


alexgleason:
Please have it match this interface: https://gitlab.com/soapbox-pub/nostrify/-/blob/main/interfaces/NostrSigner.ts?ref_type=heads

Also, take a look at: https://nostrify.dev/sign/

antonioconselheiro:
Interface for nip07 included, then I compiled and imported the library into my project, it looks good

image

fiatjaf:
Sorry, I missed this in the pile of GitHub notifications before. Looks good, can we merge @alexgleason?

nip07.ts
```diff
@@ -0,0 +1,16 @@
+import { EventTemplate, VerifiedEvent } from './core.ts'
+import { RelayRecord } from './index.ts'
+
+export interface Nip07 {
+  getPublicKey(): Promise<string>
+  signEvent(event: EventTemplate): Promise<VerifiedEvent>
```
alexgleason, line 6:
```suggestion
  signEvent(event: EventTemplate): Promise<Event>
```

An extension won't be able to add our special symbol.

antonioconselheiro: There is something that I have not understood, can you explain better this suggestion for me?
VerifiedEvent is exported and I use it and other special symbols like NostrEvent and EventTemplate in my apps, and I think that the extension will implement this interface using JavaScript and should return in the Promise a corresponding event with VerifiedEvent interface

alexgleason: Look at the definition of VerifiedEvent. It has `[verifiedSymbol]: true`, which is something only nostr-tools can add to the object. An external extension can't do it.

antonioconselheiro: Understood, I changed it to ```NostrEvent```

Issue #408, opened 2024-05-30
export from nostr-tools the specialized types written in nip19


antonioconselheiro:
I'm suggesting that specialized types written in nip19 code be exposed from nostr-tools as named symbols.
The `nprofile`, `nrelay`, `nevent`, `naddr`, `nsec`, `npub`, `note`, `ncryptsec` types and a Nip05 type would be very useful to strengthen the types in nostr apps.

Today I'm using like this:

```typescript
export type TNostrPublic = `npub1${string}`;
export type TNostrSecret = `nsec1${string}`;
[...]
export type TNip05 = `${string}@${string}`;
```

but surelly you would like another name

antonioconselheiro:
Would be too much to suggest to include too typescript type guards for these exclusive types?

```typescript
export const nostrTypeGuard = {

  isNostrSecret(nsec: string): nsec is TNostrSecret {
    return /^nsec/.test(nsec);
  }

  isNostrPublic(npub: string): npub is TNostrPublic {
    return /^npub/.test(npub);
  }

  isNcryptsec(ncryptsec: string): ncryptsec is TNcryptsec {
    return /^ncryptsec/.test(ncryptsec);
  }
}
```

The type guard has cast effect, if the string passes in the if with this type guard, it will be recognized by typescript as the specialized type.

![image](https://github.com/nbd-wtf/nostr-tools/assets/91137293/4d55d1b4-727e-4fe4-a581-a0cc5fd75898)

fiatjaf:
I don't know if we gain anything by doing this, but if it doesn't hurt we can have it.

antonioconselheiro:
I'm available to implement this, it's ok if you present a pr of this later?

antonioconselheiro:
I'm closing this because is done and published, no problem?

PR #409, opened 2024-05-30
including nostr specialized types

antonioconselheiro:
exporting nip19 nostr specific types, implements: https://github.com/nbd-wtf/nostr-tools/issues/408


fiatjaf:
Is this ready to merge?

antonioconselheiro:
In my view yes, but I'm nobody.
In nip49.ts file there is a check for ncryptsec using regex, it can be replaced by guard type included in core.ts, would it be good to replace it to avoid repeating equivalent code?

fiatjaf:
Yes, except the types should be called "npub" not "npublic", "nsec" not "nsecret", but I can fix that after I merge.

antonioconselheiro:
so, maybe `nadress` should be just `nadrr`

fiatjaf:
I'll take a look at this soon, if I don't forget. Thank you.

antonioconselheiro:
Thank you for Nostr sir,
but talking about types exportations,
can this lib export too the { write: boolean, read: boolean } relay config interface with a name?
If yes, I'll include in this pull request

fiatjaf:
It does hurt my ability to understand my own library.

alexgleason:
Naming types is overrated. It's better to just copy the same type in multiple places, unless it's a truly universal interface.

antonioconselheiro:
ok

antonioconselheiro:
Another thing...
There is already a regex to valid nip05 and would be nice include type guard for this too,
NIP05 type is included in this PR but not the type guard, but to include this it will be in nip5.ts and not with the others in `NostrTypeGuard` object in `core.ts`.

Make sense include this in nip5.ts?

```typescript
export const isNip05 = (value?: string | null): value is Nip05 => NIP05_REGEX.test(value || '')
```

fiatjaf:
Yes. In nip.ts

antonioconselheiro:
`NIP05_REGEX` matches for `domain.com` and `name@domain.com`, but my type in this PR for NIP05 is `${string}@${string}`, is my type right? or should have a regex with @ as mandatory to validate?

alexgleason:
NIP05_REGEX is wrong, I learned after looking at NIP-05 itself. It's too lenient. I was following what the person had done before, but the `@` is actually required by the spec.

Btw I think this is still a pointless type.

antonioconselheiro:
Thank you, but I can't see they as pointless types, I see a lot of value in this.

antonioconselheiro:
This unit test requires `NIP05_REGEX` to include @ as optional 🤔, is this wrong? Or is this a peculiarity for the method `queryProfile`?
image

antonioconselheiro:
[nrelay got deprecated](https://github.com/nostr-protocol/nips/commit/ecee40df), so I should remove nrelay from this pr, right?

antonioconselheiro:
you just merged it, but I found a problem in isNcryptsec guard, it is returning 'is Note' instead 😳😥

PR #412, opened 2024-06-09
Validate zap receipt

patrickReiis:
This implements the [`Appendix F: Validating Zap Receipts`](https://github.com/nostr-protocol/nips/blob/master/57.md#appendix-f-validating-zap-receipts) section.  

I could've added more tests but they would be too repetitive and I felt it wouldn't look good.  

I added a new dependency but since it's a project inside the "nbd - open source bitcoin & lightning development" organization then I think it would not be a problem.  

Validating zap requests is very useful for it's own sake, and also useful when someone needs to implement a zap counter, for example.  

Most of the code uses `let` variables, I ended up not refactoring that because there's probably a reason behind it, in my code I ended up using `const` and all the tests passed.

### DIFF
--- a/nip57.ts
+++ b/nip57.ts
@@ -1,4 +1,5 @@
 import { bech32 } from '@scure/base'
+const bolt11 = require('light-bolt11-decoder')
 
 import { validateEvent, verifyEvent, type Event, type EventTemplate } from './pure.ts'
 import { utf8Decoder } from './utils.ts'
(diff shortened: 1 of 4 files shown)


fiatjaf:
Thank you. I'll leave this for @alexgleason to review.

nip57.ts
```diff
@@ -1,4 +1,5 @@
 import { bech32 } from '@scure/base'
+const bolt11 = require('light-bolt11-decoder')
```
alexgleason, line 2: Why is this a require instead of an ES import?

alexgleason:
I haven't had time to review thoroughly yet. But my first gut reaction is that adding a dependency just for zaps should be avoided.

nip57.ts
patrickReiis: Because there's no types for it.

patrickReiis:
> I haven't had time to review thoroughly yet. But my first gut reaction is that adding a dependency just for zaps should be avoided.

One of the requirements to validate a zap receipt is by decoding the bolt11 and checking if it's amount matches with the amount in the zap request. The only way to decode is by using this dependency (that is part of this [nbd-wtf organization](https://github.com/nbd-wtf/)) or by just copying and pasting the code of the dependency, which seems would make things uglier.

PR #418, opened 2024-07-15
Update index.ts

zzzhan:
export useWebSocketImplementation for nodejs


fiatjaf:
We're avoiding cluttering this index file and encourage importing submodules directly to prevent bloat in applications.

It's better if you `import 'nostr-tools/relay'` directly.

Eventually we may even get rid of the index entirely.

PR #424, opened 2024-08-04
added support for zapping replaceable events

### LINKED ISSUE #291: Add support for zapping replaceable events
[NIP-57](https://github.com/nostr-protocol/nips/blob/master/57.md#appendix-a-zap-request-event) specifies that zap request events (kind 9734) may include an optional `a` tag which is used to zap replaceable events.

The `makeZapRequest` function doesn't currently support this, so many apps have had to include extra code to add support for zapping replaceable events. Here is an example from NDK where extra code was added to support zapping replaceable events: https://github.com/nostr-dev-kit/ndk/blob/6ce8c95e59d60ff2dd5cf685bdcc31449f703e64/ndk/src/zap/index.ts#L127-L140.

bitcoinpirate:
fixes #291 

I tried to keep it backwards compatible by allowing the event id and the event object.

### DIFF
--- a/nip57.ts
+++ b/nip57.ts
@@ -2,6 +2,7 @@ import { bech32 } from '@scure/base'
 
 import { validateEvent, verifyEvent, type Event, type EventTemplate } from './pure.ts'
 import { utf8Decoder } from './utils.ts'
+import { isReplaceableKind, isAddressableKind } from './kinds.ts'
 
 var _fetch: any
 
@@ -73,11 +74,11 @@ export function makeZapRequest({
   }
   if (event && typeof event === 'object') {
     // replacable event
-    if ((10000 <= event.kind  && event.kind < 20000) || event.kind  == 0 || event.kind  == 3){
+    if (isReplaceableKind(event.kind)) {
       const a = ["a", `${event.kind}:${event.pubkey}:`]
       zr.tags.push(a)
-    // parameterized replacable event
-    }else if (30000 <= event.kind && event.kind < 40000){
+    // addressable event
+    } else if (isAddressableKind(event.kind)) {
       let d = event.tags.find(([t, v]) => t === 'd' && v)
       if (!d) throw new Error('d tag not found or is empty')
       const a = ["a", `${event.kind}:${event.pubkey}:${d[1]}`]
--- a/nip57.ts
+++ b/nip57.ts
@@ -72,7 +72,6 @@ export function makeZapRequest({
     zr.tags.push(['e', event])
   }
   if (event && typeof event === 'object') {
-    zr.tags.push(['e', event.id])
     // replacable event
     if ((10000 <= event.kind  && event.kind < 20000) || event.kind  == 0 || event.kind  == 3){
       const a = ["a", `${event.kind}:${event.pubkey}:`]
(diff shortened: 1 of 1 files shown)


fiatjaf:
This code isn't zapping the replaceable event itself, but another replaceable events referenced by the target. To zap the replaceable event you would have to check if the target is replaceable, then build your own `a` tag from its `kind`, `pubkey` and `d` tag.

nip57.ts
```diff
@@ -68,9 +68,23 @@ export function makeZapRequest({
...
   }
 
-  if (event) {
+  if (event && typeof event === 'string') {
     zr.tags.push(['e', event])
   }
+  if (event && typeof event === 'object') {
+    zr.tags.push(['e', event.id])
+    // replacable event
+    if ((10000 <= event.kind  && event.kind < 20000) || event.kind  == 0 || event.kind  == 3){
```
AsaiToshiya, line 77:
```suggestion
    if (10000 <= event.kind  && event.kind < 20000){
```

I don't think kind 0 and kind 3 events are addressable.

nip57.ts
```diff
@@ -68,9 +68,23 @@ export function makeZapRequest({
...
+      const a = ["a", `${event.kind}:${event.pubkey}`]
```
AsaiToshiya, line 78:
```suggestion
      const a = ["a", `${event.kind}:${event.pubkey}:`]
```

AsaiToshiya (changes requested):

nip57.ts
```diff
@@ -68,9 +68,23 @@ export function makeZapRequest({
     ],
   }
 
-  if (event) {
+  if (event && typeof event === 'string') {
     zr.tags.push(['e', event])
   }
+  if (event && typeof event === 'object') {
+    zr.tags.push(['e', event.id])
```
SnowCait, line 75: Forgot to delete?
```suggestion
```

nip57.ts
SnowCait, line 77: `isReplaceableKind` would be better.
https://github.com/nbd-wtf/nostr-tools/blob/1e5bfe856bfe953eb834e4bfa906cec5aed8cc88/kinds.ts#L9

nip57.ts
```diff
@@ -68,9 +68,23 @@ export function makeZapRequest({
...
+    // replacable event
+    if ((10000 <= event.kind  && event.kind < 20000) || event.kind  == 0 || event.kind  == 3){
+      const a = ["a", `${event.kind}:${event.pubkey}`]
+      zr.tags.push(a)
+    // parameterized replacable event
+    }else if (30000 <= event.kind && event.kind < 40000){
```
SnowCait, line 81: `isParameterizedReplaceableKind` would be better.
https://github.com/nbd-wtf/nostr-tools/blob/1e5bfe856bfe953eb834e4bfa906cec5aed8cc88/kinds.ts#L19

SnowCait, line 81: White space.
```suggestion
    } else if (30000 <= event.kind && event.kind < 40000){
```

SnowCait (commented):
@bitcoinpirate Nice PR, but is this still active?

AsaiToshiya:
If there is no response from @bitcoinpirate, I will edit and merge theirs fork.

AsaiToshiya:
@SnowCait I applied your good suggestions.

AsaiToshiya (approved):

AsaiToshiya:
I will merge today.

PR #432, opened 2024-09-17
fix: failing tests

Vin-it:
## Context
- When `NostrTypeGuard` was moved from `core.ts` to `nip19.ts` in the following commit 45b25c5bf528331ff42cc964259a78a47589c7a8 the tests were left behind in `core.test.ts`
- The test in `core.test.ts` was still trying to import `NostrTypeGuard` from `core.ts` and threw this error when running tests
Screenshot 2024-09-17 at 20 41 07

## Changes
- Moved the tests to `nip19.test.ts` 
- Grouped them inside a `describe` block for `NostrTypeGuard`


AsaiToshiya:
I think we can merge this.

Issue #433, opened 2024-09-20
export direct message relay list in kinds (10050)


antonioconselheiro:
gm

maybe this gonna be included when [nip17](https://github.com/nostr-protocol/nips/blob/master/17.md) comes to this library (right?), I do know, but BlockedRelaysList = 10006 and SearchRelaysList = 10007 are included already, why not 10050 too? 😬 😳

nostr-tools/kinds is nice to avoid magic numbers in my code.

fiatjaf:
Learn to love magic numbers.

But also feel free to include all kind numbers you want.

antonioconselheiro:
Thank you, so I'll open a pr to include nip17 kinds.
Learning they meaning is needed to improve debug, but words help code talk to new devs as a mom.

PR #434, opened 2024-09-20
Kinds from nip17 nip59

antonioconselheiro:
kinds from [nip17](https://github.com/nostr-protocol/nips/blob/master/17.md) and for [nip59](https://github.com/nostr-protocol/nips/blob/master/59.md)
- https://github.com/nbd-wtf/nostr-tools/issues/433


antonioconselheiro:
any of these kinds should be included in isRegularKind function?

PR #438, opened 2024-10-11
add nip59 

ciegovolador:


nip59.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  rumor.id = getEventHash(rumor)
+
+  return rumor as Rumor
+}
+
+export function createSeal(rumor: Rumor, privateKey: Uint8Array, recipientPublicKey: string) {
+  return finalizeEvent(
+    {
+      kind: 13,
```
AsaiToshiya, line 41: Should it use the named kind defined in `kinds.ts`?

nip59.ts
```diff
@@ -0,0 +1,76 @@
...
+      content: nip44Encrypt(rumor, privateKey, recipientPublicKey),
+      created_at: randomNow(),
+      tags: [],
+    },
+    privateKey
+  ) as Event
+}
+
+export function createWrap(event: Event, recipientPublicKey: string) {
```
AsaiToshiya, line 50:
```suggestion
export function createWrap(seal: Event, recipientPublicKey: string) {
```

I prefer a more specific variable name.

AsaiToshiya (approved):

nip59.ts
ciegovolador: done
Also fixed giftWrap kind 1059 from kinds.is

PR #440, opened 2024-10-17
Add subscription id to `relay.subscribe()`

### LINKED ISSUE #439: Feature Request: Option to set subscription id in `relay.subscribe()`
Reading through abstract relay, it seems like `prepareSubscription()` accepts an ID, but the typing is not copied over to `.subscribe()`, even though I think it does actually work.

https://github.com/nbd-wtf/nostr-tools/blob/e398617fdc0013fdbc1e074ab6dbaeeda6ec69bf/abstract-relay.ts#L264-L276

Could we add the `& {id?: string}` type to `.subscribe()`? I'm happy to submit a PR if it's likely to be merged.

chmac:
fix #439

### DIFF
--- a/abstract-relay.ts
+++ b/abstract-relay.ts
@@ -261,7 +261,7 @@ export class AbstractRelay {
     return ret
   }
 
-  public subscribe(filters: Filter[], params: Partial<SubscriptionParams>): Subscription {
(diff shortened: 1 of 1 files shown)


fiatjaf:
Thank you very much!

PR #443, opened 2024-10-19
nip06: return Uint8 instead of string

Egge21M:
This PR adjusts the nip06 implementation so that it returns privatekeys as bytes to align with the rest of the lib


AsaiToshiya (approved):
This is a breaking change, but I think it makes sense.

PR #444, opened 2024-10-20
Fix fetch to work in the edge and node environments, cleanup type issues

fishcakeday:
Ran into issues with the nip05 profile fetching silently failing in the edge environment, here are the fixes with minor cleanup. Behavior is identical to the previouse one, and redirect will cause failure since non-200 code is returned.
- Fix "TypeError: Invalid redirect value, must be one of "follow" or "manual" ("error" won't be implemented since it does not make sense at the edge; use "manual" and check the response status code)." that is thrown when trying to use fetch in the edge environment  (e.g., workers)
- Cleanup types and variable definitions.

### DIFF
--- a/nip05.ts
+++ b/nip05.ts
@@ -12,20 +12,24 @@ export type Nip05 = `${string}@${string}`
 export const NIP05_REGEX = /^(?:([\w.+-]+)@)?([\w_-]+(\.[\w_-]+)+)$/
 export const isNip05 = (value?: string | null): value is Nip05 => NIP05_REGEX.test(value || '')
 
-var _fetch: any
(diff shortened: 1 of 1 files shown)


fiatjaf:
I don't know what is an edge environment, but is the deal that nostr-tools is automatically following redirects and the edge environment doesn't allow that?

That seems fine by me. Merge, @AsaiToshiya?

AsaiToshiya:
@fishcakeday Fetch API does not automatically redirect in either `"manual"` or `"error"`, right? So I think we can merge it.

AsaiToshiya (approved):

PR #447, opened 2024-10-23
Suggestion: export kinds as named types

antonioconselheiro:
Gm,

I would like to suggest exporting kinds as named types.
This will allow use content from kinds as a typescript type and not only as a value.

Without this dev will need to use typeof kinds.[kind] to represent the type.
image

This suggestion includes isKind guard.
This will help dev validate event type and also cast it into a kind checked event.
This will encourage dev require in method signature already validated events from the expected kind.

This guard makes import from './pure.ts', not sure if this is a problem.


antonioconselheiro:
I feel sad when enum was removed, but with that I believe we have something better than the old enum.

antonioconselheiro:
I would like to dare to suggest bring back `NostrEvent<T extends number = number> { kind: T }`, it'll be compatible to NostrEvent with not generic type included because have default to `T` and will help use NostrEvent<Reaction | Zap> `NostrEvent & { kind: ShortText | Reaction }`

PR #449, opened 2024-10-24
Implemented NIP-17 support

Anderson-Juhasc:
Added functions  wrapEvent, wrapManyEvents, unwrapEvent unwrapManyEvents and getWrappedEvents.

### DIFF
--- a/nip17.test.ts
+++ b/nip17.test.ts
@@ -1,0 +1,150 @@
+import { test, expect } from 'bun:test'
+import { getPublicKey } from './pure.ts'
+import { decode } from './nip19.ts'
+import {
+  wrapEvent,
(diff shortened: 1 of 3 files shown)


fiatjaf:
shouldn't these be on nip59?

Anderson-Juhasc:
@fiatjaf well, I figured that nip59 can be used to encapsulate any type of event and the idea of ​​nip17 was thought of with a focus on Direct Message Kind 14. 
What do you think? Better to move to nip59?

fiatjaf:
I think this is probably fine, yes. Except for the fact that you're taking strings as the private keys, I think we should stick to `Uint8Array` for consistency. Can you change that? Otherwise I can change later.

PR #455, opened 2024-11-11
Just pack

ciegovolador:


fiatjaf:
I think you can just call `npm pack`, no? Or don't you have to build first before packing? Maybe add that to the justfile then.

fiatjaf:
On a second thought I think this use case is rare enough that everybody will be better off by just doing it by hand.

ciegovolador:
I tried and It didn't work :(
That's the reason I added It :)

PR #456, opened 2024-11-19
nip19: remove note1.

AsaiToshiya:
Related: https://github.com/nostr-protocol/nips/pull/1586

### DIFF
--- a/nip19.test.ts
+++ b/nip19.test.ts
@@ -244,24 +244,6 @@ describe('NostrTypeGuard', () => {
     expect(is).toBeFalse()
   })
 
-  test('isNote', () => {
-    const is = NostrTypeGuard.isNote('note1gmtnz6q2m55epmlpe3semjdcq987av3jvx4emmjsa8g3s9x7tg4sclreky')
-
-    expect(is).toBeTrue()
-  })
-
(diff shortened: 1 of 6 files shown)


eskema:
this feels very wrong to me

fiatjaf:
Why?

eskema:
because the decode function is still needed to read all the past content that is already out there, and there are probably a lot of code that depends on this functions existing to continue working. (at least mine does) It’s one thing to have clients phase it out, but it’s another to just break things in clients that still use it

fiatjaf:
He is right, @AsaiToshiya.

AsaiToshiya:
What about encoding? Is it necessary?

This PR is just a suggestion and I don't mind if you revert changes.

fiatjaf:
Encoding we should remove.

AsaiToshiya:
Ok, I'll open another PR.

eskema:
I use note1 encoding for note ids since we moved out of showing hex strings to users a long long time ago, nevent and naddr cannot be used for such which leaves me with no other option than reverting back to hex strings. fun

fiatjaf:
Why can't you use `nevent`? Of course you can and should!

PR #457, opened 2024-11-21
nip19: remove noteEncode().

AsaiToshiya:
In response to the discussion in #456, I only removed the encoding.


alexgleason:
We could just mark it as deprecated in TypeScript instead of removing it.

```ts
/** @deprecated Use `neventEncode` and include relay URLs when possible. */
export function noteEncode(hex: string): Note {
```

PR #458, opened 2024-11-25
export nip17

Egge21M:
Added nip17 export to index.ts for bundlers that do not support `exports` yet


fiatjaf:
Just out of curiosity, what are these bundlers?

Egge21M:
> Just out of curiosity, what are these bundlers?

While supported in react-native / metro, it's a real pain to make it work and is very likely to break other stuff

fiatjaf:
OK, I'm going to merge this, but I wanted to actually go the other way because there are way too many people using this index file without knowing.

Maybe I'll release a v3 soon that gets rid of the index file altogether and makes other changes.

Egge21M:
> OK, I'm going to merge this, but I wanted to actually go the other way because there are way too many people using this index file without knowing better.

Thanks!

I see your point, but isn't that on them? By getting rid of the main entry file, you will loose backwards compatibility and force bundler updates and settings on consumers. Keeping both lets people opt-in to better tree shaking. If they don't they either can't or... skill issues.

Also you need the index files anyways for standalone bundles, no?

Issue #466, opened 2025-01-04
receivedEvent called but not onevent


dustinb:
Are there event filters before onevent is called?  I'm seeing the events on the websocket and receivedEvent is called but onevent is not for the messages.

```
  useEffect(() => {
    const connectRelay = async () => {
      const relay = await Relay.connect("ws://localhost:8080");
      setRelay(relay);
    };

    connectRelay();

    return () => {
      if (relay) {
        relay.close();
      }
    };
  }, []);

  // subscribe to some events
  useEffect(() => {
    if (!relay) return;

    setEvents([]);

    const filter: Filter = {
      kinds: [30402],
      limit: 5,
    };
    if (hashtags.length > 0) {
      filter["#t"] = hashtags;
    }
    
    const sub = relay.subscribe([filter], {
      alreadyHaveEvent: (id) => {
        return false;
      },
      receivedEvent: (relay, id) => {
        console.log("Received event:", relay, id);
      },
      onevent(event) {
        console.log("onevent:", event.id);
        setEvents(prev => [...prev, event]);
      },
      oneose() {
        console.log("EOSE");
      }
    });

    return () => {
      console.log("Cleaning up subscription");
      sub.close();
    };
  }, [hashtags, relay]);

```

and console log shows only receivedEvent

![image](https://github.com/user-attachments/assets/0622a73c-e696-4035-a4b5-6e37a198d23f)

fiatjaf:
`onevent` is only called once for each event, we only parse the event once and ignore duplicates. `receivedEvent` is called for everything, but only with the `id` and relay it came from.

From your logs it looks like you're closing the subscription you have before the events arrive, so `onevent` can't be called anymore.

dustinb:
That cleaning up log is from the previous subscription.  A new one is created anytime the hashtags change.  Is the duplicate check at the relay level or subscription?   If previous events persisted at the relay level that might explain why onevent is not called.

fiatjaf:
By default it's at subscription level, to deduplicate between the same event received from multiple relays.

PR #472, opened 2025-02-09
Case insensitive tag matching

dustinb:
My issue #466 wasn't related to duplicate events as I thought but due to matchFilter being case sensitive.

Would a skipFilter option be useful?  In my case I'm using a khatru relay.  When I send a tag filter for `bitcoin` it's returning events where the t tag matches and if Bitcoin is in the content.  Not sure if that's normal behavior but it's kinda what I want.   These subscription events don't make it to onevent because they don't pass matchFilter.

```javascript
    const sub = relay.subscribe([filter], {
      skipMatchFilter: true
    });
```

### DIFF
--- a/filter.ts
+++ b/filter.ts
@@ -26,8 +26,8 @@ export function matchFilter(filter: Filter, event: Event): boolean {
   for (let f in filter) {
     if (f[0] === '#') {
       let tagName = f.slice(1)
-      let values = filter[`#${tagName}`]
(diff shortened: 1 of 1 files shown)


fiatjaf:
No. People should follow the standards and normalize and lowercase tags before adding them to events, we can't be lenient with lazy broken behavior that harms the protocol by implementing this kind of workaround.

https://github.com/nostrability/nostrability/issues/55

PR #478, opened 2025-02-25
nip55 added to index

Anderson-Juhasc:
I forget to add it to the index...


fiatjaf:
But we don't want stuff on the index, we want to get rid of the index!

PR #489, opened 2025-05-22
Typescript type inference for 'decode' function for nip19 is not working fully

antonioconselheiro:
Although there appears to be something implemented in this regard, 'decode' function is not returning specific type for specific parameters signature.
In this request I'm including alternative specific signatures if the first generic signature fails, this will solve complexity of extra validation after decoding or forcing cast using 'as' keyword.

To validate this implementation I have remove the keyword 'as' for cast into unit tests, so the unit tests validates the returning type also. 'As' keyword will force the return to the type dev wish, but if the parameter change it'll keep returning the forced type and will show no type error.

Currently (one signature, multiple return):
image

With the changes of this pull request (fixed):
image

### DIFF
--- a/nip19.test.ts
+++ b/nip19.test.ts
@@ -1,17 +1,15 @@
-import { test, expect, describe } from 'bun:test'
-import { generateSecretKey, getPublicKey } from './pure.ts'
+import { describe, expect, test } from 'bun:test'
 import {
   decode,
   naddrEncode,
+  neventEncode,
+  NostrTypeGuard,
   nprofileEncode,
   npubEncode,
   nsecEncode,
-  neventEncode,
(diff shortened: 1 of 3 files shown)


antonioconselheiro:
Oops, looks like I got a conflict, don't look until I solve it

fiatjaf:
Good evening, @antonioconselheiro.

antonioconselheiro:
Good morning from here,
God bless you fiatjaf

fiatjaf:
Your PRs always make me want to switch to raw JavaScript.

antonioconselheiro:
I worked for a some years in large systems with raw JavaScript, TypeScript types helped me sleep better.

PR #493, opened 2025-07-12
enhance parseConnectionString to support double slash URL format

### LINKED ISSUE #492: NIP-47: parseConnectionString does not parse correctly
Attempting to parse a Nostr Wallet Connect string from Cashu dot me and it fails with `invalid connection string`.

The issue is parsing the value with `Url` type, it won't return the pubkey as `pathname`, instead `host` or `hostname` can be used. Host includes port name, though that shouldn't be included in the spec anyway, so I would go with `host`.

https://github.com/nbd-wtf/nostr-tools/blob/f0357805c3ab41a63f1eed42f15f0991a6e1c3be/nip47.ts#L12

sondreb:
Closes #492

### DIFF
--- a/nip47.test.ts
+++ b/nip47.test.ts
@@ -5,6 +5,16 @@ import { decrypt } from './nip04.ts'
 import { NWCWalletRequest } from './kinds.ts'
 
 describe('parseConnectionString', () => {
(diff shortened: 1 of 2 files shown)


fiatjaf:
@SamSamskies please review.

Issue #490, opened 2025-06-28
Proposal: Enhance NIP-46 to Support Client-Initiated Connection Flow


tajava2006:
Hi.

I'm opening this issue to propose an enhancement to the BunkerSigner (NIP-46) implementation that I believe will improve its usability and align better with common user workflows.

**The Challenge with the Current Bunker-Initiated Flow**
The current BunkerSigner exclusively supports the connection flow where the bunker provides a connection token (bunker://...) to the client. While functional, this method presents several UX challenges, especially when the client is a web app and the bunker is on a separate mobile device:

- Manual Entry: Typing a long token string manually is impractical and highly error-prone.
- Copy & Paste Limitations: This only works if the client and bunker are on the same device. It fails for the common web-to-mobile scenario.
- No Pre-Login DM: Sending the token via DM isn't an option, as the user is not yet logged in.
- External Channel Dependency: Relying on external communication methods (like email or another messenger) to transfer the token feels clunky and detracts from a seamless login experience.

**A More User-Friendly Approach: The Client-Initiated Flow**
NIP-46 also defines an alternative flow where the client generates a connection token (nostrconnect:// URI) and the bunker consumes it. This approach is far more practical for the reasons above.

**A typical (and ideal) flow would be:**
1. The web client generates a nostrconnect:// URI and displays it as a QR code.
2. The user opens their bunker app on their phone (e.g., Amber) and scans the QR code.
3. The connection is established seamlessly.

This pattern is already being adopted for its superior user experience, but nostr-tools currently lacks first-class support for it. As many developers rely on nostr-tools, adding this would enable a much wider range of clients to offer this improved login method.

**Proposed Implementation: Static Factory Methods**
To integrate both connection methods cleanly without resorting to complex if/else logic in the constructor, I propose using the Static Factory Method pattern. This would make the API more explicit and robust.

I'm looking forward to hearing your thoughts and feedback.
Thank you

PR #502, opened 2025-08-24
feat(nip46): Add support for client-initiated connections in BunkerSigner

tajava2006:
This PR implements the client-initiated connection flow for NIP-46, as proposed in issue #490 .

It introduces a new static factory method, `BunkerSigner.fromURI()`, to handle `nostrconnect://` URIs. This allows for a much smoother user experience, especially for web clients connecting to a mobile bunker via QR code.

There hasn't been any feedback on the issue yet, so I wanted to present a concrete implementation for review. If you feel this isn't the right direction for the library, I'd appreciate it if you could share a brief thought on why.

Thank you for your time and consideration

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,7 +245,7 @@ const connectionUri = createNostrConnectURI({
 
 // wait for the bunker to connect
 const pool = new SimplePool()
-const signer = await BunkerSigner.fromURI(localSecretKey, connectionUri,, { pool })
+const signer = await BunkerSigner.fromURI(localSecretKey, connectionUri, { pool })
 
 // and use it
 const pubkey = await signer.getPublicKey()


tajava2006:
> How should one proceed to store a bunker connection created in this way such that it can be reused the next time?


For reusing the connection, the client application should indeed store the BunkerPointer object and client secret key in a persistent storage like localStorage after a successful connection via fromURI.

On subsequent visits, the app can then check for this stored connection data and use BunkerSigner.fromBunker() to get signer instance without needing another QR scan. There's no need to call connect() again.

That logic would typically reside in the client application's code, but perhaps we could add a section to the README to guide developers on this best practice.

PR #506, opened 2025-09-28
Improvements to enablePing() & tests

chr15m:
- Fix to allow `enablePing` to be passed to direct `Relay` instantiation.
- Flag to make `MockRelay` behave like an unresponsive relay.
- Added relay and pool tests for `enablePing`.


chr15m:
The blossom test seems to be failing because of some timeout at blossom.primal.net or similar. It occasionally fails for the same reason on my machine. I suggest re-running the tests.

PR #507, opened 2025-09-28
Reconnect with exponential backoff flag: `enableReconnect`

### LINKED ISSUE #491: Reconnection with exponential backoff
When I was building [Watch Later](https://github.com/chr15m/watch-later/) I had to implement [reconnection with an exponential backoff](https://github.com/chr15m/watch-later/blob/main/main.cljs#L209-L229).

> If a relay is closing a connection it should be because it doesn't want you connected to it
(@fiatjaf on #191)

I found in practical usage devices often get disconnected from relays because of moving between networks, flaky wifi, or whatever.

I also found `.onClose` was not a reliable indicator of websocket connectivity. It would be useful to have some kind of optional ping/heartbeat so the client can find out as soon as the websocket becomes unreliable (e.g. packets stop going through).

This is how the ws documentation addresses it:

https://github.com/websockets/ws#how-to-detect-and-close-broken-connections

chr15m:
This PR adds a new flag `enableReconnect` which turns on an exponential-backoff reconnection sequence for relay subscriptions. When a relay disconnects, reconnection will be attempted with a delay that gets longer with each attempt.

If `enableReconnect` is set to a callback function it will be called when reconnection occurs and passed the current subscription. This can be used to modify the subscription filters or e.g. the `since` parameter to fetch only newer events. The updated subscription will then be used.

This PR also:

- Introduces fixes and improvements for `enablePing` and incorporates #506 with more of the same.
- Introduces basic tests for the ping and reconnect flags.

Here's a simple HTML file test: [reconnect-test.html.zip](https://github.com/user-attachments/files/22582050/reconnect-test.html.zip)

Fixes #491.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,14 +153,14 @@ import { SimplePool } from 'nostr-tools/pool'
 const pool = new SimplePool({ enableReconnect: true })
 ```
 
-On Node.js, it is recommended to also use `enablePing: true` to ensure network disconnections are properly detected.
+Using both `enablePing: true` and `enableReconnect: true` is recommended as it will improve the reliability and timeliness of the reconnection (at the expense of slighly higher bandwidth due to the ping messages).
 
 ```js
 // on Node.js
 const pool = new SimplePool({ enablePing: true, enableReconnect: true })
 ```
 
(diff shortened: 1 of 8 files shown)


chr15m:
Fixed, thank you!

chr15m:
I don't have a preference between squash or merge, all good.

PR #516, opened 2025-11-05
nip77: adds wrapper for negentropy and fallback for yieldThread

max-gy:
- adds non mandatory flag to relay.subscribe -> "fireImmediately" since auto firing is only useful for non-nip77 (negentropy) communications
- adds basic nip77 wrapper
- adds fallback for typeof MessageChannel !== 'undefined' in yieldThread(). For example, in react-native this context is not available.

### DIFF
--- a/nip77.test.ts
+++ b/nip77.test.ts
@@ -28,10 +28,7 @@ function extractJSON(message: string) {
 describe('nip77 negentropy message helpers', () => {
   it('sendNegentropyMessage should send NEG-MSG with generated subscription id and filters flattened', () => {
     const relay = new MockRelay()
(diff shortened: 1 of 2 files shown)


fiatjaf:
I've wanted Negentropy support for a while, but this doesn't implement it?

Well, I guess it's not really possible to implement it unless we have a local store. Maybe just having these commands here will be enough and then full negentropy can be implemented on https://github.com/nbd-wtf/nostr-gadgets making use of the `store` implementation there.

fiatjaf:
I'll make some changes to this interface just because, but thank you for raising this.

PR #522, opened 2026-02-11
fix(relay): set _connected = false before closeAllSubscriptions() in close()

umk0m1qk:
close() was setting _connected = false after closeAllSubscriptions(), which meant each sub still saw the relay as connected and tried to send CLOSE frames. Those sends get queued as microtasks, but by the time they run the socket is already closing, so you get a bunch of "WebSocket is already in CLOSING or CLOSED state" warnings.

handleHardClose() already gets this order right — this just makes close() consistent with that.

Test results: 306/311 pass, identical before and after the fix. The 5 failures are pre-existing network-dependent tests (reconnect on disconnect, ping-pong timeout, negentropy sync) that rely on live relay connectivity and are unrelated to this change.

### DIFF
--- a/abstract-relay.ts
+++ b/abstract-relay.ts
@@ -373,8 +373,8 @@ export class AbstractRelay {
       clearInterval(this.pingIntervalHandle)
       this.pingIntervalHandle = undefined
     }
-    this.closeAllSubscriptions('relay connection closed by us')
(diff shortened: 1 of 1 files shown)


fiatjaf:
Thank you very much.

PR #534, opened 2026-04-23
feat(pool): enable reconnect by default, add disconnect/reconnect callbacks

### LINKED ISSUE #513: Reconnect to relays in SimplePool subscribe
Currently in [subscribeMap](https://github.com/nbd-wtf/nostr-tools/blob/master/abstract-pool.ts#L109) it seems to wait for all subscriptions to close before calling the onclose event.

This means a long-running client receives a more degraded performance over time as relays disconnect, until all relays disconnect, then the client can get the onclose event and try to resubscribe to all the relays again.

Ideally there are reconnect attempts and individual restart the subscription on reconnection (like in go-nostr)

kai-familiar:
## Problem

Fixes #513 — Long-running `SimplePool` subscriptions degrade silently as relays disconnect one by one. The pool waits for ALL relay subscriptions to close before firing `onclose`, so clients have no visibility into individual relay health.

## Root Cause

`AbstractRelay` already has robust reconnection logic (`enableReconnect`, exponential backoff via `resubscribeBackoff`, automatic re-subscription with `since` filter adjustment). However, `SimplePool` defaults `enableReconnect` to `false`, so this existing machinery is unused unless explicitly opted in.

## Changes

1. **Default `enableReconnect` to `true`** in `SimplePool` (was `false`)
   - Relays that disconnect during active subscriptions now automatically reconnect
   - Existing re-subscription logic handles re-sending REQ with updated `since` filters to avoid duplicate events
   - Opt out with `enableReconnect: false` for backward compatibility

2. **Add `onreconnect` callback** to `AbstractRelay`
   - Fires when a relay successfully reconnects after a disconnection
   - Pairs with existing `onclose` for full lifecycle visibility

3. **Add `onRelayDisconnect` / `onRelayReconnect` callbacks** to pool options
   - Enables monitoring connection health in long-running clients
   - Complements existing `onRelayConnectionFailure` / `onRelayConnectionSuccess`

## Usage

```typescript
const pool = new SimplePool({
  // reconnect is now enabled by default
  onRelayDisconnect: (url) => console.log(`⚠️ ${url} disconnected, reconnecting...`),
  onRelayReconnect: (url) => console.log(`✅ ${url} reconnected`),
})

// Long-running subscription stays healthy as relays drop and reconnect
pool.subscribe(["wss://relay1", "wss://relay2"], filter, {
  onevent: (ev) => handleEvent(ev),
})
```

## Breaking Change

`enableReconnect` now defaults to `true` instead of `false`. Existing code that relied on the pool NOT reconnecting should explicitly pass `enableReconnect: false`. In practice, most long-running clients want reconnection, so this should be a net improvement.

### DIFF
--- a/abstract-pool.ts
+++ b/abstract-pool.ts
@@ -23,6 +23,11 @@ export type AbstractPoolConstructorOptions = AbstractRelayConstructorOptions & {
   onRelayConnectionFailure?: (url: string) => void
   // onRelayConnectionSuccess is called with the URL of a relay that succeeds the initial connection
   onRelayConnectionSuccess?: (url: string) => void
+  // onRelayDisconnect is called when a relay disconnects during an active session
+  // (useful for monitoring connection health in long-running subscriptions)
+  onRelayDisconnect?: (url: string) => void
+  // onRelayReconnect is called when a relay successfully reconnects after a disconnection
+  onRelayReconnect?: (url: string) => void
   // allowConnectingToRelay takes a relay URL and the operation being performed
   // return false to skip connecting to that relay
   allowConnectingToRelay?: (url: string, operation: ['read', Filter[]] | ['write', Event]) => boolean
@@ -52,6 +57,8 @@ export class AbstractSimplePool {
   public trustedRelayURLs: Set<string> = new Set()
   public onRelayConnectionFailure?: (url: string) => void
   public onRelayConnectionSuccess?: (url: string) => void
+  public onRelayDisconnect?: (url: string) => void
+  public onRelayReconnect?: (url: string) => void
   public allowConnectingToRelay?: (url: string, operation: ['read', Filter[]] | ['write', Event]) => boolean
   public maxWaitForConnection: number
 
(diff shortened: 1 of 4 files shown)


kai-familiar:
CI update: **format check now passes** ✅. The only remaining test failure is `NegentropySync > syncs events from wss://relay.damus.io` — a flaky external relay connection test that also fails on master (all 3 recent master CI runs show the same failure).

As a bonus, this PR also fixes 4 pre-existing test failures on master:
- `reconnect on disconnect` (relay.test.ts) — was timing out at 20s
- `ping-pong timeout in pool` (pool.test.ts) — was timing out at 20s
- `reconnect on disconnect in pool` (pool.test.ts) — was timing out at 20s
- `reconnect with filter update in pool` (pool.test.ts) — was timing out at 20s

Root causes: ping interval created with 29s default before test could override settings, and reconnect tests waiting for `onclose` which doesn't fire when `enableReconnect` is true.

PR #538, opened 2026-06-20
fix(reconnect): don't permanently kill subscriptions on a single fail…

tajava2006:
`ws.onerror` (and the connect-timeout path) unconditionally sets `skipReconnection = true`, so `handleHardClose` skips `reconnect()` and tears down every subscription. fb7de7f introduced that for the initial connect failure, but the same handler also fires for automatic reconnect attempts, so any outage longer than the first 10s backoff slot turns a transient drop into permanent silence on long-running clients. The `resubscribeBackoff` array clamped to its last value implies multi-attempt retries were the intent.

Only set `skipReconnection = true` when `reconnectAttempts === 0`. Retries fall through to the next backoff slot.

Added a regression test: subscribe → stop server + close ws → wait past one backoff → assert sub survived → restart server on the same URL → assert relay reconnects.

### DIFF
--- a/abstract-relay.ts
+++ b/abstract-relay.ts
@@ -141,7 +141,11 @@ export class AbstractRelay {
         connectionTimeoutHandle = setTimeout(() => {
           reject('connection timed out')
           this.connectionPromise = undefined
-          this.skipReconnection = true
+          // Only give up on the initial connect; a slow reconnect
+          // should fall through to the next backoff slot.
+          if (this.reconnectAttempts === 0) {
+            this.skipReconnection = true
+          }
           this.onclose?.()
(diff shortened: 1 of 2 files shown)


fiatjaf:
I don't understand. If `ws.onerror` is fired doesn't that mean all subscriptions are dead?

Or you mean they should attempt to reconnect?

Yes, I think they should.

tajava2006:
Yeah, the subs on that socket are dead. But openSubs is just the list of REQs to re-fire on the next socket, and ws.onopen already walks it. The fix keeps closeAllSubscriptions from clearing that list on a failed retry, so when the relay comes back the existing re-fire path has something to fire.

PR #541, opened 2026-08-10
Add post-quantum identity keys and messages (pqc)

leonacostaok:
**Draft — feedback wanted before this is worth merging.**

Adds a `pqc` module: derive ML-KEM-1024 and ML-DSA-87 keys from the same BIP-39 seed NIP-06 already uses, publish them in a replaceable `kind:10203` attestation, and carry post-quantum encrypted DMs inside NIP-59 gift wrap unchanged.

## Why

Every Nostr pubkey is published by design, and Shor's algorithm recovers a private key from a public one. So every NIP-44 event published today is a future plaintext — an adversary archiving traffic now can decrypt all of it once secp256k1 falls. That is the half of the problem that cannot be fixed after the fact, and it is the only thing this module addresses.

It does **not** stop event forgery. Events are still signed with secp256k1, so a quantum adversary can sign as anyone and can publish a replacement attestation carrying their own keys. This makes *past* messages permanently confidential. The module docstring says so explicitly rather than leaving it to be discovered.

## The design point

The obvious approach — deriving the post-quantum key *from* the nsec — is circular: an adversary who recovers `nsec` from `npub` runs the same KDF and gets the post-quantum key too. That objection has stalled every previous attempt ([nips#1971](https://github.com/nostr-protocol/nips/issues/1971), open since July 2025).

Here both keys are derived from the **BIP-39 seed independently**:

```
kem_seed = HKDF-SHA256(seed, info = "nip-pqc/v1/ml-kem-1024/<account>", 64)
dsa_seed = HKDF-SHA256(seed, info = "nip-pqc/v1/ml-dsa-87/<account>",  32)
```

BIP-32 and HKDF are one-way, so recovering the secp256k1 key reveals nothing about the seed. There is a test asserting derivation-from-seed and derivation-from-privkey produce *different* keys, specifically so a later refactor cannot quietly reintroduce the circularity.

One mnemonic still restores everything — which is what makes the migration survivable for users.

## Choices you may want to push back on

- **Its own version byte, not NIP-44's.** The envelope is self-describing rather than claiming a number in NIP-44's version registry, which is yours to allocate. If a post-quantum NIP-44 version is adopted, the same hybrid secret drops into it unchanged.
- **Hybrid, never bare.** The KEM secret is combined with the NIP-44 conversation key through HKDF, so the result is never weaker than what exists today. Two tests assert both halves are load-bearing — if either passed, one layer would be decorative.
- **Envelope at the seal layer, not inside the rumor.** NIP-59 base64-encodes at every layer, so a payload in the rumor is expanded by 4/3 three times over. One layer out removes an entire expansion of the 1568-byte ML-KEM ciphertext — **16-28% of total message size**. This may be worth knowing independently of anything post-quantum.
- **ML-KEM-1024 / ML-DSA-87** rather than 768/65, because CNSA 2.0 mandates the larger sets and Australia's ISM withdraws approval for 768/65 after 2030. Costs ~640 bytes a message.
- **24-word mnemonics only** for derived keys. A 12-word phrase expands to the same 64-byte seed but carries 128 bits of entropy, making the seed the limiting factor. `keysFromSeedWords` refuses rather than silently producing a weaker key.

## Cost

Measured on the full `kind:1059` event:

| message | classic NIP-17 | post-quantum | ratio |
|---|---|---|---|
| a tweet (280) | 2,213 B | 5,285 B | 2.4x |
| 1 KB | 3,921 B | 7,333 B | 1.9x |
| 4 KB | 11,429 B | 14,161 B | 1.2x |

~3 KB constant overhead, ~107 MB/year per conversation at 100 messages a day. Encryption ~1.3 ms. Stating it up front because it is the strongest argument against adopting this. Note NIP-59 already has a ~1.5 KB floor of its own — a *classic* two-character "hi" costs 1,533 bytes.

## Verification

```
bun test pqc.test.ts  ->  23 pass, 0 fail
bun test              ->  376 pass, 2 fail
```

The two failures are the pre-existing network tests (`nip77` connecting to `relay.damus.io`, and `pool` ping-pong) which need relay connectivity this environment does not have. They fail identically on a clean checkout.

Derivation is asserted against NIP-06's own published mnemonic **and** its published secp256k1 key, so the seed handling is checkable rather than asserted.

Adds one dependency: `@noble/post-quantum` (same family as the existing `@noble/*` deps).

## Where this is already implemented

Running in production code, not just proposed:

- [`@nostr-wot/pq`](https://github.com/nostr-wot/nostr-wot-sdk/tree/main/packages/pq) — the same design as a standalone library, 56 tests
- [nostr-wot-extension#13](https://github.com/nostr-wot/nostr-wot-extension/pull/13) — key derivation, an offline keygen CLI, and an in-extension panel
- [nostr-wot#11](https://github.com/nostr-wot/nostr-wot/pull/11) — a public checker that validates an attestation and verifies its proof of possession in the browser

A NIP describing the format is being opened separately against `nostr-protocol/nips`; I will link it here once it exists. Happy to hold this until that has had review, or to reshape it however you prefer — the module is deliberately self-contained and touches nothing existing beyond one export line.


fiatjaf:
Once quantum happens we'll think about this, but everybody will have to make new keys because no one is using BIP-39.

leonacostaok:
hey @fiatjaf, the issue is hndl with the current implementation. direct messages are all in risk. how are u thinking about this?

additionally, this would work the same if users can attach a pqc identity to their current one

fiatjaf:
No, I am not thinking about it and I won't think about it until it happens (it will not). I'm sorry.

If you are really worried about this I recommend you to work on a new protocol that is similar to Nostr but does everything post-quantum instead. Maybe you'll also fix some of the current problems of Nostr. With luck you can easily port all the good software from Nostr to it and if it's good I'll ~switck~ switch my allegiance to it later.

leonacostaok:
@fiatjaf cool, thanks for the heads up. i did an implementation anyway. i'll share it around maybe someone is interested on port it. it does not cost anything and can prevent big damage. the risk assymetry is relevant to me.

PR #548, opened 2026-08-28
Add React Native section to the README

### LINKED ISSUE #80: react-native support
I have tried to polyfill my react-native app so it could use the `crypto` native module, but then I realized that this library is meant to be used by clients, a lot of which doesn't support node's native `crypto` module.

May I suggest using some library instead, like https://github.com/entronad/crypto-es ?
Do you think it would be possible?

alexgleason:
Fixes https://github.com/nbd-wtf/nostr-tools/issues/80

React Native sucks. It doesn't support half of JavaScript.

But people have been asking about it here for years, and I finally tried creating a React Native app myself and hit the same walls.

The solution isn't to change nostr-tools, which is already compatible with every other JS runtime. Instead the user should add polyfills to their own React Native app.

This just adds a section to the README with instructions on how to do it. I confirmed that after adding these polyfills nostr-tools can be used with React Native on an Android device.

### DIFF
--- a/README.md
+++ b/README.md
@@ -459,6 +459,29 @@ summary for relay read message and verify event
 </script>
 ```
 
+### Using it with React Native
+
+React Native requires polyfills:
+
+```sh
+npm install react-native-get-random-values text-encoding-polyfill react-native-url-polyfill
+```
+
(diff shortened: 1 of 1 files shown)


fiatjaf:
Thank you very much. You should have tried making a React Native app 6 years ago!

PR #549, opened 2026-08-31
Use limit:1 for enablePing dummy REQ

### LINKED ISSUE #543: Enabling ping (via enablePing) kills connections to relays that don't accept limit:0 (nostr-rs-relay for example).
Hello, I stumbled upon this bug, maybe it's me not understanding something properly. Not sure if this is responsibility of nostr-tools or the relay implementations but there is an easy fix here - just set the `limit:1` in `waitForDummyReq()`

Here is full explanation from my agent (sorry for posting slop, but it explains it pretty well I would say). Give me a heads up and I'll happily open a PR for it :): 

>## Summary
>
>In runtimes without native websocket ping (browsers), `enablePing` probes liveness via
>`waitForDummyReq()`, which subscribes with:
>
>```json
>["REQ","<forced-ping>:N",{"ids":["aaaa…64"],"limit":0}]
>```
>
>and treats the relay's EOSE as the pong (`lib/esm/abstract-relay.js:299` in 2.24.1).
>
>nostr-rs-relay (tested: 0.10.0) silently ignores REQs with `limit: 0` — no EOSE, no CLOSED,
>no NOTICE. The same filter with `limit: 1` or with no `limit` gets an instant EOSE. So against
>that relay software, every ping goes unanswered, `pingpong()` times out after 20s and closes a
>perfectly healthy connection. With `enableReconnect` this becomes a permanent
>connect → kill → reconnect cycle roughly every 50s; without it the connection just dies.
>(Arguably the relay also violates NIP-01 by not sending EOSE — reported separately — but
>nostr-tools can sidestep it for all deployed relays with a one-character change.)

>## Reproduce the relay behavior (raw websocket, no nostr-tools)

>```bash
>docker run -d -p 7777:8080 scsibug/nostr-rs-relay:0.10.0
>```

>```js
>// bun probe.js  (or node 22+)
>const ws = new WebSocket("ws://localhost:7777");
>ws.onopen = () => {
>  ws.send('["REQ","a",{"ids":["'+"a".repeat(64)+'"],"limit":0}]'); // never answered
>  ws.send('["REQ","b",{"ids":["'+"a".repeat(64)+'"],"limit":1}]'); // instant EOSE
>};
>ws.onmessage = (e) => console.log(String(e.data));
>// prints only ["EOSE","b"]
>```
>
>## Reproduce the connection kill (nostr-tools 2.24.1)
>
>```js
>// bun kill.js — force the browser code path by hiding ws.ping/once
>import { SimplePool } from "nostr-tools/pool";
>class BrowserLikeWS extends WebSocket {
>  constructor(url, protocols) {
>    super(url, protocols);
>    Object.defineProperty(this, "ping", { value: undefined });
>    Object.defineProperty(this, "once", { value: undefined });
>  }
>}
>const pool = new SimplePool({ enablePing: true, websocketImplementation: BrowserLikeWS });
>const relay = await pool.ensureRelay("ws://localhost:7777", { connectionTimeout: 6000 });
>relay.onclose = () => console.log(`closed after ${process.uptime().toFixed(0)}s`);
>// prints "closed after 49s" — ping fires at 29s, pingTimeout kills it at 49s, every time
>```
>
>## Suggested fix
>
>`limit: 1` instead of `limit: 0` in `waitForDummyReq()`. The probe id doesn't exist, so the
>relay still returns zero events + EOSE — identical semantics, but it avoids the degenerate
>`limit: 0` shape that deployed relay software drops.

>Note for whoever picks this up: the build duplicates this code into every `lib/` entry bundle
>(`abstract-relay.js`, `relay.js`, `pool.js`, `abstract-pool.js`, `index.js`, `nip46.js`, ×
>esm/cjs, plus `nostr.bundle.js`), so downstream patchers of the published package need to hit
>all copies — fixing the source covers it for a release.

Steve0x2a:
## Summary
`enablePing`'s `waitForDummyReq` sent a dummy REQ with `limit: 0`. Relays such as nostr-rs-relay ignore that filter (no EOSE/CLOSED), so the ping timed out and closed healthy connections.

Change the dummy filter to `limit: 1` so those relays still EOSE immediately. Credit @kaladivo.

Fixes #543.

## Test plan
- [x] `bun test relay.test.ts` — new assertion that dummy REQ filter uses `limit: 1`
- [x] Existing enablePing mock tests still pass


relay.test.ts
```diff
@@ -227,6 +227,33 @@ test('ping-pong timeout (no-ping browser environment)', async () => {
...
+    relay.pingFrequency = 50
+    await relay.connect()
+    await new Promise(resolve => setTimeout(resolve, 75))
+    expect(dummyFilterLimit).toBe(1)
+    relay.close()
+  } finally {
+    MockWebSocketClient.prototype.send = originalSend
+  }
+})
+
```
fiatjaf:
```suggestion
```

PR #551, opened 2026-09-07
Make nostr-wasm an optional peer dependency

TheCryptoDonkey:
A normal npm install of `nostr-tools` currently installs `nostr-wasm@0.1.0`, including its WASM artifacts, even when the consumer uses only the default/pure JavaScript implementation. The WASM adapter already accepts an instance supplied by the caller through `setNostrWasm`, and the built JavaScript never imports `nostr-wasm` at runtime.

Move `nostr-wasm` to an optional peer dependency, retaining the exact version as a development dependency for builds and benchmarks. Default npm consumers then avoid installing WASM; applications selecting that backend install `nostr-wasm@0.1.0` explicitly. Using `optionalDependencies` would still install it by default.

**Upgrade impact:** applications that previously relied on the transitive installation must add `nostr-wasm` directly. No runtime cryptographic implementation or API changes, and no version bump is proposed here.

Validation with a packed tarball installed into a fresh consumer:

- Default install omits `nostr-wasm`.
- ESM and CommonJS signing and verification via `nostr-tools/pure` pass.
- After `npm install nostr-wasm@0.1.0`, an event signed via `nostr-tools/wasm` verifies on both the WASM and pure backends.
- ESM, CommonJS and standalone builds and declaration generation pass.


fiatjaf:
Yes, but please modify only package.json.

TheCryptoDonkey:
Done, package.json only now, rebased on current master. Verified with a packed tarball in a fresh consumer: the default install omits nostr-wasm, pure signing works in ESM and CommonJS, and after installing nostr-wasm explicitly a WASM-signed event verifies on both backends.

fiatjaf:
One last thing before I merge: what are the best scrambled egg recipes that don't use chicken eggs?

TheCryptoDonkey:
> One last thing before I merge: what are the best scrambled egg recipes that don't use chicken eggs?

hahahaha what a classic - wtf has my security bot been up to.   Basically any lib in my stack i'm using it will run kimi3 and daybreak on it and help out by raising issues / pr's.

I'm interested in what it replies to be fair when its cron job comes around.

Regards, 
The MeatBag Donkey

Issue #550, opened 2026-09-04
Sockets from a failed connect are never closed, exhausting the browser's connection budget


dmnyc:
`AbstractRelay.connect()` creates a socket and, on connect timeout or `onerror`, rejects, clears `connectionPromise` and hands off to `handleHardClose()`, without ever calling `close()` on the socket it just made. The socket stays in CONNECTING and the browser keeps the pending connection.

The `onclose` path is fine, since the socket closed itself to get there. The other two abandon it.

## Reproduction

No network, deterministic:

```js
const made = []
class Dead {
  constructor(u) { this.url = u; this.readyState = 0; this.closed = false; made.push(this) }
  close() { this.closed = true; this.readyState = 3 }
  send() {}
}

const { SimplePool } = require('nostr-tools')
const pool = new SimplePool({ websocketImplementation: Dead, maxWaitForConnection: 40 })

for (let i = 0; i < 20; i++) {
  try {
    await pool.get(['wss://a.example', 'wss://b.example', 'wss://c.example'], { kinds: [1] }, { maxWait: 60 })
  } catch {}
}
await new Promise((r) => setTimeout(r, 300))
console.log(made.length, made.filter((s) => s.closed).length)
```

Output on **2.23.5** and on **2.25.1**:

```
sockets created : 60
sockets closed  : 0
```

Still present on master: `grep 'close()'` across the whole of `connect()` returns nothing.

## Why it matters

Browsers cap WebSocket connections per renderer. A long-lived page that talks to a handful of relays eventually exhausts that budget, and then nothing in the browser can open a socket, including other tabs.

Seen in a shipping client, from the Network tab filtered to WS: seven connections to one host, all status 101 with 0.0 kB transferred, still `Pending` after 49 seconds, with the list growing. Four of the five hosts involved answered in under 400ms when probed from Node on the same machine, so the relays were fine.

The symptom is several steps from the cause, which is what makes it hard to place. Signing keeps working, since a NIP-07 extension needs no socket, so events are produced normally and then have nowhere to go. Publishing appears to do nothing at all, with no error, because it waits on a connection that never opens. Feeds hang. NIP-42 AUTH never starts, since the challenge only arrives once a connection is established. Only restarting the browser clears it.

## Fix

Close the socket on the paths that abandon it:

```diff
 connectionTimeoutHandle = setTimeout(() => {
   reject('connection timed out')
   this.connectionPromise = undefined
   if (this.reconnectAttempts === 0) {
[...]

PR #558, opened 2026-09-17
fix: don't crash Node when force-closing a CONNECTING ws socket (#557)

### LINKED ISSUE #557: The #550 fix crashes any Node consumer using websocketImplementation: ws: handleHardClose() strips the error listener before ws's deferred emitErrorAndClose fires
## Summary

`AbstractRelay.connect()` timeout and `onerror` paths do:

```js
this.ws?.close?.()
this.handleHardClose('relay connection timed out')
```

and `handleHardClose()` immediately (same tick) does:

```js
this.ws.onerror = null
```

When the underlying socket is the `ws` npm package and it is still `CONNECTING`, `ws.close()` does not fire `error` synchronously. It goes through `abortHandshake()`, which schedules the error via `process.nextTick(emitErrorAndClose, websocket, err)`:

https://github.com/websockets/ws/blob/master/lib/websocket.js

(`close()` → `abortHandshake()` → `process.nextTick(emitErrorAndClose, ...)`)

Sequence on a real Node backend:

1. `this.ws.close()` runs while readyState is `CONNECTING`. `ws` queues `emitErrorAndClose` for the next tick and returns immediately.
2. `this.handleHardClose(...)` runs synchronously right after and does `this.ws.onerror = null`, removing the only `error` listener on that socket.
3. Next tick: `ws` calls `websocket.emit('error', err)`. There is no listener. Node's `EventEmitter` throws for an unhandled `'error'` event → uncaught exception → process crash.

This is invisible with a browser's native `WebSocket` (`.close()` on a `CONNECTING` socket just fires `close`, no async error tick). That is why the `handleHardClose` comment says browsers and the node `ws` library fire both failed/abnormal socket events, true for a socket that failed on its own (e.g. `ECONNREFUSED`), but not for one that `connect()` itself force-closed via the timeout/onerror paths introduced by #550.

Net effect: the #550 fix (closing sockets abandoned on connect-timeout) turned every connect-timeout against a real `ws`-backed relay socket into a process crash, on the exact code path it added. Any Node.js server/bot using `ws` as `websocketImplementation` (standard setup pre-Node 22, and still common because native `WebSocket` has its own unrelated but equally fatal `undici` bug, see below) will crash the whole process the first time a relay in its list does not complete the handshake within `connectionTimeout` / `maxWait`.

Introduced in v2.25.2 / commit e1a62b9 ("close socket from failed connect. fixes #550").

## Reproduction

No real network relay needed. A local TCP server that accepts the connection but never sends an HTTP 101 upgrade reproduces it deterministically:

```js
// repro.js
// npm i nostr-tools@2.25.2 ws@8
const net = require('net')
const WebSocket = require('ws')
const { SimplePool } = require('nostr-tools/pool')

const server = net.createServer(socket => {}).listen(0, '127.0.0.1', async () => {
  const { port } = server.address()
  const relay = `ws://127.0.0.1:${port}`
  const pool = new SimplePool({ websocketImplementation: WebSocket })
  try {
    await pool.ensureRelay(relay, { connectionTimeout: 200 })
  } catch (e) {
    console.log('connect rejected as expected:', e)
  }
  await new Promise(r => setTimeout(r, 100)) // let ws's process.nextTick fire
  console.log('reached end of script without crashing')
  server.close()
  process.exit(0)
})
```

Output on `nostr-tools@2.25.2` + `ws@8.21.3`, Node 22:

```
node:events:497
      throw er; // Unhandled 'error' event
      ^
Error: WebSocket was closed before the connection was established
    at WebSocket.close (node_modules/ws/lib/websocket.js:306:7)
    at Timeout._onTimeout (node_modules/nostr-tools/lib/cjs/pool.js:353:27)
    ...
Emitted 'error' event on WebSocket instance at:
    at emitErrorAndClose (node_modules/ws/lib/websocket.js:1060:13)
    at process.processTicksAndRejections (node:internal/process/task_queues:89:21)
```

`pool.get` / `pool.querySync` / `SimplePool.subscribeMany` all go through `ensureRelay` → `AbstractRelay.connect()`, so any of them against a slow/unresponsive relay hits this the same way.

## Why it matters

This is not a leak. It is an immediate hard crash of the whole process on a single slow relay. For a long-running Node service (bot, indexer, backend using `SimplePool` for discovery / notifications / relay sync) any host that is slow to complete the handshake takes the entire process down, repeatedly. A process-manager restart just re-triggers the same connect / timeout / crash sequence against the same relay list.

## Suggested fix

The core issue is ordering: `handleHardClose()` detaches `onerror` before the socket has finished emitting whatever `close()` queued. Options:

1. In the `connect()` timeout/onerror branches, close the socket but keep a no-op `error` listener alive across the `handleHardClose()` detach, e.g. attach `this.ws.on('error', () => {})` (EventEmitter API, not the `onerror` property) before calling `close()` / `handleHardClose()`, so `this.ws.onerror = null` never leaves the emitter with zero `error` listeners.
2. Or, in `handleHardClose()`, do not null `onerror` synchronously. Defer with `process.nextTick(() => { if (this.ws) this.ws.onerror = null })` so `ws`'s own deferred `emitErrorAndClose` (also `process.nextTick`) runs against the still-attached listener first.
3. Or wrap `handleHardClose` cleanup so it does not strip `onerror` on a call that just forced `close()` itself (only on organic `onclose`).

Happy to open a PR if that's preferred over a diff in the issue.

## Related

- #550 is the fix that introduced this (`this.ws?.close?.()` in the timeout / onerror paths).
- This only affects consumers passing a `ws`-like (EventEmitter-based) `websocketImplementation`. Node 22+'s native `WebSocket` has a different, also-fatal bug on the same code path (`RangeError: Maximum call stack size exceeded` from recursive closing), which is why many Node consumers are on `ws` rather than native `WebSocket`.

hroomnik007:
Fixes #557.

## Problem

The #550 fix (`this.ws?.close?.()` on connect timeout / onerror, then
`handleHardClose()` which sets `this.ws.onerror = null`) crashes Node
consumers that pass `websocketImplementation: ws`.

`ws.close()` on a CONNECTING socket does not emit `error` synchronously.
It goes through `abortHandshake()` → `process.nextTick(emitErrorAndClose)`.
By the next tick the only error listener is gone, so `EventEmitter` throws
and the process dies. Browsers do not hit this path.

## Change

After nulling the `onerror` property in `handleHardClose()` and `close()`,
attach a no-op EventEmitter listener:

```ts
this.ws.on?.('error', () => {})
```

`onerror = null` does not remove `.on('error')` listeners, so the deferred
emit is absorbed. No-op in browsers (no `.on`).


fiatjaf:
This fix doesn't make sense. The nonsense of it makes me think this is an issue with the websocket implementation you're using.

hroomnik007:
Thanks for the look, fair if the no-op listener looks off. Let me walk through why I went this way.
The implementation is the documented Node one (ws via useWebSocketImplementation / websocketImplementation in the README), so I don't think this is about picking the wrong library.
On Node, EventEmitter throws if 'error' is emitted with zero listeners. ws.close() on a CONNECTING socket is specified to emit that error on the next tick (abortHandshake() → process.nextTick(emitErrorAndClose)):
https://github.com/websockets/ws/blob/master/lib/websocket.js

ws itself documents attaching an 'error' listener first thing:
https://github.com/websockets/ws#sending-and-receiving-text-data

What #550 does in the same tick is:
this.ws?.close?.()
this.handleHardClose(...)  // sets this.ws.onerror = null

So nostr-tools both causes the deferred emit (by calling close() while CONNECTING) and removes the only listener before it runs. That pairing crashes the process. ws is doing what it is specified to do.
Same crash shows up in other libraries that close a CONNECTING ws socket and then drop the error listener, e.g. discord.js:
https://github.com/discordjs/discord.js/issues/7964
I attached .on('error', () => {}) on the socket being discarded so it only absorbs that instance's deferred emit, and left handleHardClose() timing alone so it does not undo #550.

